PR #29, opened 2017-05-23
make mod cluster public

vthriller:


jhecking:
You are right, that the `scan_node`/`query_node` methods are currently unusable. However, rather than making the cluster mod public, I would rather add an interface to the `Client` to get a list of the active cluster nodes. I have filed a new issue #33 and will close this PR. I will add the required functionality in a separate, new PR.

May I ask what your use case is and whether you are actively building a new application using the Rust client or whether you are just testing it out?

jhecking:
New PR filed: #34

vthriller:
> May I ask what your use case is and whether you are actively building a new application using the Rust client or whether you are just testing it out?

I was just playing aroung with both Rust and Aerospike, although to be more specific, in this particular case I was just looking for a way to retrieve statistics without spawning `asinfo -v namespace/whatever` and parsing its output.

jhecking:
> I was just playing aroung with both Rust and Aerospike, although to be more specific, in this particular case I was just looking for a way to retrieve statistics without spawning asinfo -v namespace/whatever and parsing its output.

And did you get this working?

vthriller:
> And did you get this working?

Unfortunately I did't have much spare time recently to figure this out.

PR #30, opened 2017-05-23
quick and tiny documentation fix

vthriller:


jhecking:
First of all: Thanks!

> version is bumped, but what about docs?

I know and I'm open to suggestions! I didn't want to set a bad example by putting something like `aerospike = "*"` in the docs. Maybe we just assume that all Rust developers are familiar with `Cargo.toml` and that it's not necessary to provide an example at all. What do you think?

vthriller:
> Maybe we just assume that all Rust developers are familiar with Cargo.toml and that it's not necessary to provide an example at all. What do you think?

I guess it's fine for pre-1.0 crate as long as users are reminded that API might be broken in the future.

(Sorry for late reply.)

Issue #47, opened 2018-04-25
Small change to put API.


baldmountain:
I'm working on an Elixir nif using Rust and rustler to access an Aerospike server from Elixir. I'm new to Rust and had an issue trying to implement a wrapper for put and needed some help. You can see the full discussion on the rust-lang forum here: https://users.rust-lang.org/t/newbie-ownership-issue/17079

My issue was trying to translate an Elixir Map into a &[&Bin] to pass to put. I needed to generate a `[Bin]` and then do:

```
let bins = bins.iter().collect::<Vec<_>>();
```

to generate the [&Bin]

They made the suggestion:

```
fn put<A: AsRef<Bin>>(..., bins: &[A]) ...
// or if a slice isn’t necessary and iterator works
fn put<I: IntoIterator<Item=A>, A: AsRef<Bin>>(iter: I) ...
```

Keeping in mind that I'm a Rust newbie, I thought I'd mention it here.

jhecking:
Hi @baldmountain, thanks for the suggestion! Very exciting to see the Rust client being used in this way. I will look into the suggested API change (and whether it might apply to other functions as well). Since it looks like you have already solved your immediate problem and since I'm currently busy working on another project, it might take me a couple of days to get to this, though.

> Keeping in mind that I'm a Rust newbie, I thought I'd mention it here.

So am I. So feedback like this is extremely valuable to me! Let me know if you run into any other issue while using the Rust client or if you have any other suggestions for how to improve it. :-)

baldmountain:
I am not the person to be asking for help. I barely know what I'm doing at this point.

Don't be afraid to ask questions on the forum I linked to above. Those guys are super helpful and friendly. Asking for help there was a very positive experience. :)

jhecking:
For the upcoming v0.3.0 release, I have updated the Put API to use `AsRef`. I have considered adopting the `IntoIterator` trait but have decided against it for now, after considering the pros and cons.

```
pub fn put<'a, 'b, A: AsRef<Bin<'b>>>(
    &self, 
    policy: &'a WritePolicy, 
    key: &'a Key, 
    bins: &'a [A]
) -> Result<()>
```

Follow-up rust-lang forum discussion: https://users.rust-lang.org/t/generic-types-with-lifetimes/20142.

PR #49, opened 2018-09-07
Use generic types to make client API more flexible

jhecking:
As per #47 and this [rust-lang forum thread](https://users.rust-lang.org/t/newbie-ownership-issue/17079) any client API that takes `&[&Bin]` could be improved by taking `&[A] where A: AsRef<Bin>` instead.


jhecking:
Follow-up discussion on the rust-lang forum: https://users.rust-lang.org/t/generic-types-with-lifetimes/20142/11.

PR #64, opened 2020-05-02
Changing version on documentation

nassor:
The current version is misleading and drives users to use an old version.

_+ fix aerospike configuration on Travis_


jhecking:
Thanks!

PR #65, opened 2020-05-04
Move to rust edition 2018

nassor:
- Move to rust edition 2018
- Removing most of `clippy` warnings.


jhecking:
Thank you for your contribution!

PR #67, opened 2020-05-23
Accept batch read response without key digest

### LINKED ISSUE #63: Accept batch read response without key digest
The current batch read response protocol sends back the digest for every record.  This information was used to verify that the batch index (also sent back for every record) pointed to the correct key entry.  Because the digest is not strictly necessary, we plan to eventually remove it from the batch response protocol.

E.g. Java client change (CLIENT-1119): https://github.com/aerospike/aerospike-client-java/commit/27a855ab6902ff634562cfe19d229a5be8dd1bbb

jlr52:
Not sure if you guys still need help, but I would love to contribute :)

Resolves #63.

### DIFF
--- a/src/commands/batch_read_command.rs
+++ b/src/commands/batch_read_command.rs
@@ -144,8 +144,7 @@ impl<'a, 'b> BatchReadCommand<'a, 'b> {
                         .batch_reads
                         .get_mut(batch_record.batch_index)
                         .expect("Invalid batch index");
-                    let record = batch_record.record;
-                    batch_read.record = record;
+                    batch_read.record = batch_record.record;
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


jhecking:
Thank you for your contribution, @jlr52!

Can you please merge master branch into your branch? I have just merged another PR (#65) which I think will help address the failing unit tests.

jlr52:
absolutely, let me merge master into my branch

jhecking:
> absolutely, let me merge master into my branch

👍 Thanks!

src/commands/batch_read_command.rs
```diff
@@ -145,12 +145,6 @@ impl<'a, 'b> BatchReadCommand<'a, 'b> {
                         .get_mut(batch_record.batch_index)
                         .expect("Invalid batch index");
                     let record = batch_record.record;
-                    if let Some(Record {
-                        key: Some(ref key), ..
-                    }) = record
-                    {
-                        assert_eq!(batch_read.key.digest, key.digest);
-                    }
                     batch_read.record = record;
```
jhecking, line 148: Seems like the temp. `record` variable is no longer needed, and this could be simplified to

    batch_read.record = batch_record.record

jacklberk: good catch, let me fix that

jlr52: good catch, let me fix that

Issue #12, opened 2017-01-23
Add APIs to wait for completion of long-running server tasks


jhecking:
Certain server tasks are initiated by the client but are asynchronously executed by the server in the background. The client should provide APIs to monitor these tasks using the info protocol and notify the application when they are completed.

* [x] [Registering a new Lua UDF module](http://www.aerospike.com/docs/guide/udf.html#registering-udfs).
* [x] Creating a [secondary index](http://www.aerospike.com/docs/guide/query.html#secondary-index).
* [ ] Executing [read-write scans](http://www.aerospike.com/docs/guide/scan.html#read-write-scan).

jhecking:
Resolved in #69.

PR #69, opened 2020-05-27
Add APIs to wait for completion of long-running server tasks

jlr52:
This is a proof of concept for adding query status api similar to https://github.com/aerospike/aerospike-client-java/blob/master/client/src/com/aerospike/client/task/IndexTask.java

I have a few questions that I listed as TODO (ie. what is the appropriate error)

TODOs before possible merge:

Add more test

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,4 +1,4 @@
-// Copyright 2015-2018 Aerospike, Inc.
+// Copyright 2015-2020 Aerospike, Inc.
 //
 // Portions may be licensed to Aerospike, Inc. under one or more contributor
 // license agreements.
--- a/src/client.rs
+++ b/src/client.rs
@@ -23,7 +23,7 @@ use std::vec::Vec;
 
 use scoped_pool::Pool;
 
-use crate::query::IndexTask;
+use crate::task::{IndexTask, RegisterTask};
 use crate::batch::BatchExecutor;
 use crate::cluster::{Cluster, Node};
 use crate::commands::{
@@ -492,7 +492,7 @@ impl Client {
         udf_body: &[u8],
         udf_name: &str,
         language: UDFLang,
-    ) -> Result<()> {
+    ) -> Result<RegisterTask> {
         let udf_body = base64::encode(udf_body);
 
         let cmd = format!(
@@ -517,7 +517,10 @@ impl Client {
             );
         }
 
-        Ok(())
+        Ok(RegisterTask::new(
+            Arc::clone(&self.cluster),
+            udf_name.to_string()
+        ))
     }
 
     /// Register a package containing user-defined functions (UDF) with the cluster. This
@@ -532,7 +535,7 @@ impl Client {
         client_path: &str,
         udf_name: &str,
         language: UDFLang,
-    ) -> Result<()> {
+    ) -> Result<RegisterTask> {
         let path = Path::new(client_path);
         let mut file = File::open(&path)?;
         let mut udf_body: Vec<u8> = vec![];
@@ -569,6 +572,7 @@ impl Client {
         function_name: &str,
         args: Option<&[Value]>,
     ) -> Result<Option<Value>> {
+
         let mut command = ExecuteUDFCommand::new(
             policy,
             self.cluster.clone(),
@@ -577,6 +581,7 @@ impl Client {
             function_name,
             args,
         );
+
         command.execute()?;
 
         let record = command.read_command.record.unwrap();
@@ -828,11 +833,11 @@ impl Client {
             index_type,
             CollectionIndexType::Default,
         )?;
-        return IndexTask::new(
+        return Ok(IndexTask::new(
             Arc::clone(&self.cluster),
             namespace.to_string(),
             index_name.to_string()
-        );
+        ));
     }
 
     /// Create a complex secondary index on a bin containing scalar, list or map values. This
--- a/src/client.rs
+++ b/src/client.rs
@@ -556,11 +556,10 @@ impl Client {
         // Sample response: {"udf-remove:filename=file_name.LUA;": "ok"}
         let response = node.info(policy.base_policy.timeout, &[&cmd])?;
 
-        if response.get(&cmd).is_some() && response.get(&cmd).unwrap() == "ok" {
-            return Ok(());
+        match response.get(&cmd).map(String::as_str) {
+            Some("ok") => return Ok(()),
+            _ => bail!("UDF Remove failed: {:?}", response),
         }
-
-        bail!("UDF Remove failed: {:?}", response)
     }
 
     /// Execute a user-defined function on the server and return the results. The function operates
--- a/src/client.rs
+++ b/src/client.rs
@@ -553,9 +553,10 @@ impl Client {
     ) -> Result<()> {
         let cmd = format!("udf-remove:filename={}.{};", udf_name, language);
         let node = self.cluster.get_random_node()?;
+        // Sample response: {"udf-remove:filename=file_name.LUA;": "ok"}
         let response = node.info(policy.base_policy.timeout, &[&cmd])?;
 
-        if response.get("ok").is_some() {
+        if response.get(&cmd).is_some() && response.get(&cmd).unwrap() == "ok" {
             return Ok(());
         }
 
--- a/src/client.rs
+++ b/src/client.rs
@@ -818,7 +818,7 @@ impl Client {
         bin_name: &str,
         index_name: &str,
         index_type: IndexType,
-    ) -> Result<(IndexTask)> {
+    ) -> Result<IndexTask> {
         self.create_complex_index(
             policy,
             namespace,
@@ -827,7 +827,7 @@ impl Client {
             index_name,
             index_type,
             CollectionIndexType::Default,
-        );
+        )?;
         return IndexTask::new(
             Arc::clone(&self.cluster),
             namespace.to_string(),
--- a/src/client.rs
+++ b/src/client.rs
@@ -23,6 +23,7 @@ use std::vec::Vec;
 
 use scoped_pool::Pool;
 
+use crate::query::IndexTask;
 use crate::batch::BatchExecutor;
 use crate::cluster::{Cluster, Node};
 use crate::commands::{
@@ -817,7 +818,7 @@ impl Client {
         bin_name: &str,
         index_name: &str,
         index_type: IndexType,
-    ) -> Result<()> {
+    ) -> Result<(IndexTask)> {
         self.create_complex_index(
             policy,
             namespace,
@@ -826,7 +827,12 @@ impl Client {
             index_name,
             index_type,
             CollectionIndexType::Default,
-        )
+        );
+        return IndexTask::new(
+            Arc::clone(&self.cluster),
+            namespace.to_string(),
+            index_name.to_string()
+        );
     }
 
     /// Create a complex secondary index on a bin containing scalar, list or map values. This
--- a/src/client.rs
+++ b/src/client.rs
@@ -23,7 +23,6 @@ use std::vec::Vec;
 
 use scoped_pool::Pool;
 
-use crate::task::{IndexTask, RegisterTask};
 use crate::batch::BatchExecutor;
 use crate::cluster::{Cluster, Node};
 use crate::commands::{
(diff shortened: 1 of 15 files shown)


src/client.rs
```diff
@@ -23,6 +23,7 @@ use std::vec::Vec;
 
 use scoped_pool::Pool;
 
+use crate::query::IndexTask;
```
jhecking, line 26: I would create a new `task` module, rather than putting this under the `query` module. Reason being, that there are other long-running tasks, like executing background scans/queries, that should probably all implement a common `Task` interface.

src/query/index_task.rs
```diff
@@ -0,0 +1,111 @@
...
+    namespace: String,
+    index_name: String
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum IndexTaskResult {
+    NotFound,
+    InProgress,
+    Complete
+}
```
jhecking, line 23: Maybe we should call this the task "status", rather than the task "result". The result of the index task, is that the index is either created successfully or that an error occurred.

src/query/index_task.rs
```diff
@@ -0,0 +1,111 @@
...
+
+
+    // TODO: enforce access of this only to Client
+    /// Initializes IndexTask from client, creation should only be expose to Client
+    pub fn new(cluster: Arc<Cluster>, namespace: String, index_name: String) -> Result<Self> {
+        Ok(IndexTask {
+        	cluster: cluster,
+        	namespace: namespace,
+            index_name: index_name
+        })
```
jhecking, line 40: Why are you wrapping the return value in a `Result` when there is no chance of the function failing?

src/query/index_task.rs
```diff
@@ -0,0 +1,111 @@
...
+
+
+    /// Query the status of index creation across all nodes
+    pub fn query_status(&self) -> Result<IndexTaskResult> {
+    	let nodes = self.cluster.nodes();
+
+    	if nodes.len() == 0 {
+    		// TODO: what error should this be
+    	    bail!(ErrorKind::ServerError(ResultCode::ServerError))
+    	}
```
jhecking, line 51: I'm not convinced that this should be an error at all! Definitely not a server error, which should only be returned if the client received an error response from the server with a specific error code.

How would this condition occur? Maybe if you issue the create index command, then close the client, and then try to query the index task? If that's the case, maybe we should introduce a new "connection closed" error, indicating that the cluster object referenced by the task is no longer connected to the server?

src/query/index_task.rs
```diff
@@ -0,0 +1,111 @@
...
+    fn build_command(namespace: String, index_name: String) -> String {
+        return format!("{}{}{}{}", "sindex/", namespace, "/", index_name);
+    }
+
+	fn parse_response(response: &str) -> Result<IndexTaskResult> {
+        match response.find(SUCCESS_PATTERN) {
+            None => {
+                match (response.find(FAIL_PATTERN_201), response.find(FAIL_PATTERN_203)) {
+                    // TODO: what error should this be?
+                    (None, None) => bail!(ErrorKind::ServerError(ResultCode::ServerError)),
```
jhecking, line 79: Maybe we could return a `BadResponse` error in this case.

jhecking:
Thanks for this contribution, @jlr52!

I think this is a good, first step towards resolving #12. What would be needed next is the equivalent of the Java client's [`Task` class], which can be broken down into two parts:
1) A common `Task` interface, that all tasks like the IndexTask would implement (`public abstract int queryStatus()` in the Java client's Task implementation).
2) The common logic to take any `Task` instance, and repeated query the tasks status until it is completed (`public final void waitTillComplete()` and its variants).

src/client.rs
jlr52: Cool, let me fix that

src/query/index_task.rs
jlr52: Sounds good.

src/query/index_task.rs
jlr52: The function using that was returning `Result` so I did that to match the type, but you are absolutely right in pointing out that its unneccessary, I am removing it.

src/query/index_task.rs
jlr52: Hmm, its definitely possible to reach this branch, so I favor having a define behavior, I like your idea of connection closed error let me try that out.

src/query/index_task.rs
jlr52: sounds good

jacklberk:
Apologies for the delay, 

I fix most of the suggestions (besides `ResultCode` due to clarification needed), but I have 2 questions

1. I realize that `ResultCode` is define in proto files, is it ok if I add more result type? 
2. `Async` is available in rust 1.39 do we want to consider moving the support to that version to support `Async` for `waitTillComplete` , might not be worth it depending on the use case, I imagine udf and secondary index creation is usually not used on the fly and would be used in an one off script.

jlr52:
Apologies for the delay,

I fix most of the suggestions (besides ResultCode due to clarification needed), but I have 2 questions

I realize that ResultCode is define in proto files, is it ok if I add custom result type?
Async is available in rust 1.39 do we want to consider moving the support to that version to support Async for waitTillComplete , might not be worth it depending on the use case, I imagine udf and secondary index creation is usually not used on the fly and would be used in an one off script.

jlr52:
Apologies for the delay,

I fix all of the suggestions. Btw Async is available in rust 1.39 do we want to consider moving the support to that version to support Async for waitTillComplete , might not be worth it depending on the use case, I imagine udf and secondary index creation is usually not used on the fly and would be used in an one off script.

src/task/index_task.rs
```diff
@@ -0,0 +1,109 @@
...
+        }
+    }
+}
+
+
+
+
+
+
+
```
jhecking, line 109: Minor thing, but can you please remove all the trailing empty lines? Some of the other files you added have the as well. Best would be to use [`rustfmt`](https://github.com/rust-lang/rustfmt) to ensure consistent formatting throughout the codebase.

src/task/index_task.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub fn new(cluster: Arc<Cluster>, namespace: String, index_name: String) -> Self {
+        IndexTask {
+        	cluster: cluster,
+        	namespace: namespace,
+            index_name: index_name
+        }
+    }
+
+    fn build_command(namespace: String, index_name: String) -> String {
+        return format!("{}{}{}{}", "sindex/", namespace, "/", index_name);
```
jhecking, line 33:
```suggestion
        return format!("sindex/{}/{}", namespace, index_name);
```

src/task/index_task.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+	fn parse_response(response: &str) -> Result<Status> {
+        match response.find(SUCCESS_PATTERN) {
+            None => {
+                match (response.find(FAIL_PATTERN_201), response.find(FAIL_PATTERN_203)) {
+                    (None, None) => bail!(ErrorKind::BadResponse(format!("Code 201 and 203 missing. Response: {}", response))),
+                    (_, _) => return Ok(Status::NotFound)
+                }
```
jhecking, line 42: We don't care about the index position of these patterns so we can use `str::contains` instead of `str::find` here:

```suggestion
                if response.contains(FAIL_PATTERN_201) || response.contains(FAIL_PATTERN_203)) {
                    return Ok(Status::NotFound)
                } else {
                    bail!(ErrorKind::BadResponse(format!("Code 201 and 203 missing. Response: {}", response))),
                }
```

src/task/index_task.rs
```diff
@@ -0,0 +1,109 @@
...
+                let percent_end = match response[percent_begin..].find(DELMITER) {
+                    None =>  bail!(ErrorKind::BadResponse(format!("delimiter missing in response. Response: {}", response))),
+                    Some(percent_end) => percent_end
+                };
+                let percent_str = &response[percent_begin..percent_begin+percent_end];
+                if percent_str.parse::<isize>().unwrap() != 100 {
+                    return Ok(Status::InProgress);
+                } else {
+                    return Ok(Status::Complete);
+                }
```
jhecking, line 56: Suggestion to improve the error handling:

```suggestion
                match percent_str.parse::<isize>() {
                    Ok(100) => return Ok(Status::Complete),
                    Ok(_) => return Ok(Status::InProgress),
                    Err(_) => bail!(ErrorKind::BadResponse(
                        "Unexpected load_pct value from server".to_string()
                    )),
                }
```

src/task/index_task.rs
```diff
@@ -0,0 +1,109 @@
...
+
+            if !response.contains_key(command) {
+                return Ok(Status::NotFound);
+            }
+
+            match IndexTask::parse_response(&response[command]) {
+                Ok(Status::NotFound) => return Ok(Status::NotFound),
+                Ok(Status::InProgress) => return Ok(Status::InProgress),
+                error => return error
+            }
```
jhecking, line 86: Can simply return the parsed response here:

```suggestion
            return IndexTask::parse_response(&response[command]);
```

src/task/task.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    /// interface for retrieving client timeout
+    fn get_timeout(&self) -> Result<Duration>;
+
+    // TODO: consider adding async and only support rust 1.39.0+
+    /// logic to wait until status query is completed
+    fn wait_till_complete(&self) -> Result<Status> {
+        let now = Instant::now();
+
+        while now.elapsed().as_secs() < self.get_timeout()?.as_secs() {
```
jhecking, line 35: The global client policy timeout applies to commands that are expected to respond immediately, e.g. simple get/put commands, etc. The same timeout should not apply to potentially long-running tasks like creating a new secondary index. If we add a timeout check here, the timeout value should be passed by the app in the `wait_till_complete` function call itself.

tests/src/task.rs
```diff
@@ -0,0 +1,112 @@
...
+    let udf_name = common::rand_str(10);
+    let udf_file_name = udf_name.clone().to_owned() + ".LUA";
+
+    let register_task = client.register_udf(
+        &WritePolicy::default(),
+        code.as_bytes(),
+        &udf_file_name,
+        UDFLang::Lua).unwrap();
+
+    assert!(matches_override!(register_task.wait_till_complete().unwrap(), Status::Complete));
```
jhecking, line 47:
```suggestion
    assert!(matches_override!(register_task.wait_till_complete(), Ok(Status::Complete)));
```

tests/src/task.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    client.remove_udf(
+        &WritePolicy::default(),
+        &udf_name,
+        UDFLang::Lua
+    ).unwrap();
+
+
+    thread::sleep(Duration::from_secs(5));
```
jhecking, line 56: Why do we need to wait for 5 seconds here?

tests/src/task.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    match register_task.wait_till_complete() {
+        Err(e) => {
+            assert!(matches_override!(e.kind(), ErrorKind::Connection(_err_string)));
+        },
+        _ => {
+            panic!("error");
+        }
+
+    }
```
jhecking, line 66: Can simplify this a bit by using `matches` on the output of `wait_till_complete` directly:

```suggestion
    assert!(matches_override!(
        register_task.wait_till_complete(),
        Err(Error(ErrorKind::Connection(_), _))
    ));
```

jhecking:
Hi @jlr52,

I've merge all of the proposed changes back into your branch, and made some additional fixes as well. From my perspective, this PR is ready to merge now.

jlr52:
@jhecking  Hey, super appreciated, I royally screwed up, life got in the way and the github alert email got buried in my inbox, and I totally forgot about this pull request, let me review it really fast, again super sorry, I glanced through your code review ask and they are all pretty reasonable ask, I just forgot about this......

jhecking:
Hey, no worries! I was also busy with other things. But found some time today and thought would be good to wrap this up.

jlr52:
Everything looks really good honestly, alot of condition becomes more readable, some redundant code got deleted, also loved the optimization in Task module.  Thanks for wrapping it up !!!!

question: should we create an issue about ```matches_override``` to get rid of the hack once we update rust version?

jhecking:
> question: should we create an issue about matches_override to get rid of the hack once we update rust version?

Sure, would be good to have that as a reminder.

Issue #70, opened 2020-07-08
Support Predicate Filtering


jonas32:
Hello,

i just started to implement the Rust Client into one of my Services.
While doing that, i found out that it is not supporting [Predicate Filtering](https://www.aerospike.com/docs/guide/predicate.html) filtering. It would be great if this Client would also support it. 
Querying for multiple secondary bins is really hacky without this.

Thanks
Jonas

jhecking:
Hi @jonas32, yes, that would be great! Honestly speaking, the fastest way to get that feature added would be to submit a pull request, if possible? I could provide guidance and feedback, but do not have the time myself to take this on end-to-end at the moment.

jonas32:
Hey @jhecking, i can try to implement it. I also dont have much time to do it. It might take some time and probably your guidance will be needed.

khaf:
Since time is at a premium for all involved, I'd caution against this for a few months, since the current predicate expressions are going to be deprecated on the server side and a new API will be released in the new few months.
All efforts on this issue until then will be wasted effort.

jonas32:
At some point it would be wasted, but on the other hand, its an important feature. For me, this library is nearly useless without it for example. I would have to use the C lib with Rust bindings. 
Im not sure if i could even finish the old API before the new API is released, but im sure the old one wont be removed for legacy reasons. And supporting both cant be bad.

khaf:
Per the current plan, the old API *will* be removed, both from the server and the clients.

jonas32:
Ah ok. That changes it.
I guess the new API is not yet documented?
Do you have any infos on how long we have to wait until the release of the new one?

khaf:
No dates, but in near future. The old APIs will be supported on the server for a short while, but there will be a clean cut on the newer releases of the clients.

rbotzer:
Hi @jonas32. When the new predicate filter is in place (targeted for version 5.2) the server will still support the current API. It will be deprecated, but not removed until a later version. This gives you many months to use the current API, if you implement it.  Your PR would be appreciated.

jonas32:
Thank you for this Information @rbotzer. 
I will try to implement it and open a PR as soon as it works.

PR #71, opened 2020-07-10
Support for Predicate Filtering

jonas32:
This is an implementation for Predicate Filtering. Related to #70 

Changes:
Predicates added to query/predexp.rs
Predicate Handling added to query/statement.rs
Modified commands/buffer.rs to add Headers and Values for predicates
Added test/example to tests/query.rs
Fixed some compiler and clippy warnings in various files.

Please review and report to me if you have any ideas how to improve it. 
Docs still need some work.

### DIFF
--- a/benches/client_server.rs
+++ b/benches/client_server.rs
@@ -21,8 +21,8 @@ extern crate rand;
 
 use aerospike::{Bins, ReadPolicy, WritePolicy};
 
-use bencher::Bencher;
 use aerospike::{as_bin, as_key};
+use bencher::Bencher;
 
 #[path = "../tests/common/mod.rs"]
 mod common;
--- a/src/commands/admin_command.rs
+++ b/src/commands/admin_command.rs
@@ -22,8 +22,8 @@ use crate::cluster::Cluster;
 use crate::errors::{ErrorKind, Result};
 use crate::net::Connection;
 use crate::net::PooledConnection;
-use crate::ResultCode;
 use crate::policy::AdminPolicy;
+use crate::ResultCode;
 
 // Commands
 const AUTHENTICATE: u8 = 0;
--- a/src/commands/batch_read_command.rs
+++ b/src/commands/batch_read_command.rs
@@ -23,7 +23,7 @@ use crate::commands::{self, Command};
 use crate::errors::{ErrorKind, Result, ResultExt};
 use crate::net::Connection;
 use crate::policy::{BatchPolicy, Policy, PolicyLike};
-use crate::{BatchRead, Record, ResultCode, Value, value};
+use crate::{value, BatchRead, Record, ResultCode, Value};
 
 struct BatchRecord {
     batch_index: usize,
--- a/src/commands/buffer.rs
+++ b/src/commands/buffer.rs
@@ -26,8 +26,8 @@ use crate::policy::{
     BatchPolicy, CommitLevel, ConsistencyLevel, GenerationPolicy, QueryPolicy, ReadPolicy,
     RecordExistsAction, ScanPolicy, WritePolicy,
 };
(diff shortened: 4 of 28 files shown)


SupaMario-ops:
lol itsa shitcode

jhecking:
Thanks so much, Jonas! First glance, this looks very good! Will take a closer look and do a proper review tomorrow.

jonas32:
While testing the predexp implementation, i actually found another problem.
When you create a normal filter with as_eq and a string value as second argument on a bin without secondary index, it will try to allocate much memory. The allocation seems to be done by the error return.
The test module gives:
`thread 'src::query::query_single_consumer' panicked at 'Error(ServerError(IndexNotFound), State { next_error: None, backtrace: InternalBacktrace { backtrace: None } })', tests/src/query.rs:95:25`
then it will crash with:
`memory allocation of 269172076445696 bytes failederror: test failed, to rerun pass '--test lib'`
The test always returns 269172076445696 bytes and my service returns always 38787614419.
Probably this is related to the amount of data in the set. The set from my service only holds 2 test records.
Its definitely not a full return of the set. The set only has a few KB.

I tried this with my predexp implementation version and without. This happens with both.

jonas32:
To reproduce it, you need to make the as_eq filter on a String bin that does not hold an Index. It does not matter if the set exists or if it holds any data. Same result.
Example:
```rust
    let mut statement = Statement::new(namespace, "test_set_with_no_index", Bins::All);
    statement.add_filter(as_eq!("non_exisitng_or_unindexed", "string_value"));
    let rs = client.query(&qpolicy, statement).unwrap();
    let mut count = 0;
    for res in &*rs {
        match res {
            Ok(rec) => {
                count += 1;
            }
            Err(err) => panic!(format!("{:?}", err)),
        }
    }
    assert_eq!(count, 10);
```
This exact code snipped fails with
`
thread 'src::query::query_single_consumer' panicked at 'Error(ServerError(IndexGeneric), State { next_error: None, backtrace: InternalBacktrace { backtrace: None } })', tests/src/query.rs:96:25
test src::query::query_single_consumer ... FAILED
memory allocation of 38779975527 bytes failederror: test failed, to rerun pass '--test lib'
`
This is what it looks like exactly in the console:
grafik

Its a pretty unstable problem. If one of the tests that run before this one fails, it wont allocate... I had the case where my server hit max index count from creating test sets. If this happens, no allocation happens... At least i could also replicate it with my own application running this exact function. It fails 100% every time.

If you put an Integer value instead of "string_value", it will not allocate. I have debugged it with valgrint and gdb. Both dont mark an issue about memleak etc. but report the high allocation value, so i guess its a parsing problem. I tried to go back the whole trace and check every buffer read step by step. If this snipped is executed, it fails. If its not executed, no problems. You can comment out the other code inside the function. As long as this 4 lines run, it will crash. (in the stream_command.rs on line 104)
```rust
 if let Err(err) = conn.read_buffer(buffer::MSG_REMAINING_HEADER_SIZE as usize) {
                warn!("Parse result error: {}", err);
                return Err(err);
            }
```
It does not make sense to me why it behaves like this. I compared every function that gets executed with the ones from the Golang Client. The only difference i found is the zlib stuff. Thats why i thought it might be the problem. Sadly, i was not able to get any docs about the server answer buffers from the Support. I havent tested yet if the same problem exists with the go client. Ill probably dig into the C client today and compare this one.

jonas32:
I will keep on investigating why it behaves like this. But ill move it to an own issue because its not related to this one so this can be merged. I think its not a critical error because it only happens under special conditions, that do not appear in production usecases.

jhecking:
Agree, let's merge this, and tackle the other issue separately.

PR #74, opened 2020-07-15
Macro rule should be replaced after updating to 1.42.0

### LINKED ISSUE #73: Macro rule should be replaced after updating to 1.42.0
In ```tests/src/task.rs1``` 

```
#[macro_export]
macro_rules! matches_override {
    ($expression:expr, $($pattern:tt)+) => {
        match $expression {
            $($pattern)+ => true,
            _ => false
        }
    }
}
```

Should be replaced with default ```match``` macro https://doc.rust-lang.org/beta/std/macro.matches.html

jonas32:
Changes to the task.rs test file to replace the old macro role. (requires > 1.42.0)
Resolves #73


jonas32:
I failed the rebase... Ignore this PR...

Issue #78, opened 2020-08-04
CDT Operations for List, Map and Bitwise


jonas32:
Hello,

currently the client only supports a limited set of operations for the custom data types and bitwise.
Is there a specific reasons only the few currently are supported or was there just no time to implement the others like list remove by value?
Im probably going to implement them, but i was curious why some are supported and some are not. 
Also the list cdt seems to be not fully implemented (return types etc)

jhecking:
Hi Jonas, the Aerospike Server has added new CDT operations at various points over the last couple of years. The Rust client only implements CDT operations that were available as of ~ Oct 2017.

jonas32:
Hey Jan, i already guessed this might be the reason. Good to know. Ill start implementing the other ones.
Are there any operations other than the missing ones from list/map and bitwise completely that are missing currently?

jhecking:
In terms of new operations, AS 4.9 introduced a completely new set of operations on the [HyperLogLog (HLL) probabilistic data type](https://www.aerospike.com/docs/guide/hyperloglog.html). Also 4.7 added a new delete operation to enable atomic read-then-delete.

There were also several enhancements to the existing list/map CDT operations, which the Rust client is missing:
* Support for optional element creation in nested CDT operations – also introduced in 4.9.
* Support for operating on lists and maps nested within lists and maps using [CDT Subcontext evaluation](https://www.aerospike.com/docs/guide/cdt-context.html) was added in 4.6.
* [`INF` and `WILDCARD` types](https://www.aerospike.com/docs/guide/cdt-ordering.html) for map and list comparators were added in 4.3.1.
* Relative range operations for CDTs, and `NO_FAIL` and `DO_PARTIAL` CDT error policies, which were added in 4.3.0.

And I may have missed some other, minor things... Quite a bit of functionality that's missing from the Rust client. Really appreciate your contributions! It's really good to see more interest in the Aerospike Rust client lately.

Also, it may make sense to break down the changes into multiple PRs, e.g. by data type (list/map/bitmap) or some other criteria that makes sense to you. That would make the review process a lot easier.

jonas32:
Yes definitely.
I already found the missing HLL compared to the Java client. The current functionality is really limited. The whole implementation differs strongly from for example Go or Java right now. Ill try to get the basic implementation of operations to work and create the first PR. Probably ill rewrite the existing part completely. That might be easier and faster than comparing line to line with Go and Java.

rbotzer:
[...]

PR #79, opened 2020-08-06
Update operations

jonas32:
Big update to the operations system. (#78 )
Including:
- Operation Contexts for nested operations (done)
- Missing List Operations (done)
- List Policys (done)
- List Order (done)
- Missing Map Operations (done)
- Bitwise Operations (done)
- HLL Operations
- Predicate Expressions (done)

CAUTION: This is a breaking change. The policy and return types for Lists require additional parameters for the cdt op functions.

### DIFF
--- a/src/operations/lists.rs
+++ b/src/operations/lists.rs
@@ -146,7 +146,6 @@ pub struct ListPolicy {
 }
 
 impl ListPolicy {
-    /// Create a new map policy given the ordering for the map and the write mode.
     pub const fn new(order: CdtListOrderType, write_mode: CdtListWriteFlags) -> Self {
         ListPolicy {
             attributes: order,
--- a/src/operations/lists.rs
+++ b/src/operations/lists.rs
@@ -31,9 +31,9 @@
 //! If an index is out of bounds, a parameter error will be returned. If a range is partially out of
 //! bounds, the valid part of the range will be returned.
 
-use operations::cdt::{CdtArgument, CdtOpType, CdtOperation};
-use operations::{Operation, OperationBin, OperationData, OperationType};
-use Value;
+use crate::operations::cdt::{CdtArgument, CdtOpType, CdtOperation};
+use crate::operations::{Operation, OperationBin, OperationData, OperationType};
+use crate::Value;
 
 /// Create list append operation. Server appends value to the end of list bin. Server returns
 /// list size.
--- a/src/operations/lists.rs
+++ b/src/operations/lists.rs
@@ -32,7 +32,7 @@
 //! bounds, the valid part of the range will be returned.
 
 use crate::operations::cdt::{CdtArgument, CdtOperation};
-use crate::operations::cdt_context::CdtContext;
+use crate::operations::cdt_context::{CdtContext, DEFAULT_CTX};
 use crate::operations::{Operation, OperationBin, OperationData, OperationType};
 use crate::Value;
 
@@ -187,15 +187,11 @@ pub fn list_order_flag(order: CdtListOrderType, pad: bool) -> u8 {
 /// Server creates list at given context level. The context is allowed to be beyond list
 /// boundaries only if pad is set to true.  In that case, nil list entries will be inserted to
 /// satisfy the context position.
-pub fn create<'a>(
-    bin: &'a str,
+pub fn create(
+    bin: &str,
     list_order: CdtListOrderType,
     pad: bool,
-    ctx: &'a [CdtContext],
-) -> Operation<'a> {
-    if ctx.is_empty() {
-        return set_order(bin, list_order, ctx);
-    }
+) -> Operation {
     let cdt_op = CdtOperation {
         op: CdtListOpType::SetType as u8,
         args: vec![
@@ -205,7 +201,7 @@ pub fn create<'a>(
     };
     Operation {
         op: OperationType::CdtWrite,
-        ctx,
+        ctx: DEFAULT_CTX,
         bin: OperationBin::Name(bin),
         data: OperationData::CdtListOp(cdt_op),
     }
@@ -232,10 +228,9 @@ pub fn set_order<'a>(
 /// Create list append operation. Server appends value to the end of list bin. Server returns
 /// list size.
 pub fn append<'a>(
-    policy: ListPolicy,
+    policy: &ListPolicy,
     bin: &'a str,
     value: &'a Value,
-    ctx: &'a [CdtContext],
 ) -> Operation<'a> {
     let mut args = vec![CdtArgument::Value(value)];
     args.push(CdtArgument::Byte(policy.attributes as u8));
@@ -247,7 +242,7 @@ pub fn append<'a>(
     };
     Operation {
         op: OperationType::CdtWrite,
-        ctx,
+        ctx: DEFAULT_CTX,
         bin: OperationBin::Name(bin),
         data: OperationData::CdtListOp(cdt_op),
     }
@@ -256,10 +251,9 @@ pub fn append<'a>(
 /// Create list append items operation. Server appends each input list item to the end of list
 /// bin. Server returns list size.
 pub fn append_items<'a>(
-    policy: ListPolicy,
+    policy: &ListPolicy,
     bin: &'a str,
     values: &'a [Value],
-    ctx: &'a [CdtContext],
 ) -> Operation<'a> {
     assert!(!values.is_empty());
 
@@ -273,7 +267,7 @@ pub fn append_items<'a>(
     };
     Operation {
         op: OperationType::CdtWrite,
-        ctx,
+        ctx: DEFAULT_CTX,
         bin: OperationBin::Name(bin),
         data: OperationData::CdtListOp(cdt_op),
     }
@@ -285,8 +279,7 @@ pub fn insert<'a>(
     bin: &'a str,
     index: i64,
     value: &'a Value,
-    policy: ListPolicy,
-    ctx: &'a [CdtContext],
+    policy: &ListPolicy,
 ) -> Operation<'a> {
     let mut args = vec![CdtArgument::Int(index), CdtArgument::Value(value)];
 
@@ -298,7 +291,7 @@ pub fn insert<'a>(
     };
     Operation {
         op: OperationType::CdtWrite,
-        ctx,
+        ctx: DEFAULT_CTX,
         bin: OperationBin::Name(bin),
         data: OperationData::CdtListOp(cdt_op),
     }
@@ -310,8 +303,7 @@ pub fn insert_items<'a>(
     bin: &'a str,
     index: i64,
     values: &'a [Value],
-    policy: ListPolicy,
-    ctx: &'a [CdtContext],
+    policy: &ListPolicy,
 ) -> Operation<'a> {
     assert!(!values.is_empty());
     let mut args = vec![CdtArgument::Int(index), CdtArgument::List(values)];
@@ -324,7 +316,7 @@ pub fn insert_items<'a>(
     };
     Operation {
         op: OperationType::CdtWrite,
-        ctx,
+        ctx: DEFAULT_CTX,
         bin: OperationBin::Name(bin),
         data: OperationData::CdtListOp(cdt_op),
     }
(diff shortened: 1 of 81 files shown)


jhecking:
Will do a full review later, but just an initial observation on the breaking interface change for all list/map operations to add a context parameter. E.g. as can be seen in the tests, an empty context parameter needs to be added to all existing operations:

    let ops = &vec![lists::size("bin", &[])];

What do you think about adding methods to set/add a context to the `CdtOperation` type itself, e.g.

    let op = lists::size("bin");
    let ctx = ctx_list_index(3);
    op.set_context(&[ctx]);

Same can be done for operations which require an optional policy as well, e.g.

    let op = lists::insert_items("bin", 1, &values);
    let policy = ListPolicy::default();
    op.set_policy(policy);

These new methods should return the `CdtOperation` instance itself, so they can be chained:

    let ctx = ctx_list_index(3);
    let op = lists::insert_items("bin", 1, &values)
      .set_policy(ListPolicy::default())
      .set_context(&[ctx]);

We could even do the same for the map return type as well, though removing that argument from the map op functions would off course in itself be a breaking change.

In general, I'm not opposed to introducing backward-incompatible changes at this early stage of the client. (I.e. until we declare the client "good enough" for a 1.0 release.) But I think introducing this kind of builder pattern makes the interface future-proof as well, such that adding additional parameters will not break the API every time.

I have been following this pattern for the Node.js client. E.g. see the [ListOperation](https://www.aerospike.com/apidocs/nodejs/module-aerospike_lists-ListOperation.html#withContext) class for an example.

jonas32:
Good idea. I also didnt like the idea of passing the empty slice every time. 
I will try to build it like that and report back if it works.

jonas32:
Im currently trying to implement the HLL Operations.
While doing that, i found out that HLLs are handled different than other returns. The Golang Client for example uses a HLLValue datatype. From my understanding, its just a Vec<u8> return and acts the same way as Value::Blob. I personally never used HLL in any way, so i dont understand how to implement it. Can you explain, why it needs to be handled different and whats the reason behind HLLValue?

jonas32:
Im done implementing everything besides HLL and the Wildcard/Inf stuff. 
Ill do HLL in a second PR.
I found some caching technique in the golang client for the encoded ops, but that will also be part of a second PR.
Some parts are breaking changes, but i think they should be or cant be done different.
For example Policies on Lists. Using a set_policy method would not be consistent to Map ops or the normal Operations.
From my side, this is ready for review/merge.

jhecking:
> Policy is set in the `CdtOperation` struct. In addition, policy is built into the args array together with other parameters that can not be identified once inside. It also looks like policy has a required fixed position in this array. Changing policy with the chain function would cause rebuilding this whole array.
> To be consistent with the normal Read/Write functions, i would probably build policy into the CDT functions, like its done for the normal basic read/write functions with WritePolicy and ReadPolicy.

I see your point. Would take quite a bit of refactoring to decouple the `Operation`/`CdtOperation` structs from the final on-wire protocol layout; might hurt performance as well. I'm fine with your current implementation.

jhecking:
> While doing that, i found out that HLLs are handled different than other returns. The Golang Client for example uses a HLLValue datatype. From my understanding, its just a Vec return and acts the same way as Value::Blob. I personally never used HLL in any way, so i dont understand how to implement it. Can you explain, why it needs to be handled different and whats the reason behind HLLValue?

Will have to look into that, as I'm not familiar with the Go client implementation.

jhecking:
What I've seen so far looks great! But I need some more time for a full review, as this is a quite large PR (3000+ lines added). Please bear with me.

jonas32:
No problem, i already expected that to be a longer review process.
For the HLL implementation:
You dont have to check it in the golang one. I just used this one as reference, because its the easiest for me to port to rust.
Java or C client as reference is also fine. The easiest would be nodejs for me, but this client seems to use a lot of FFI bindings to the C lib.

jhecking:
> The easiest would be nodejs for me, but this client seems to use a lot of FFI bindings to the C lib.

Yes, the Node.js client – which I also maintain – is just a wrapper around the C client. And the C client does not have a separate type for HLL values. I've reached out to @BrianNichols, who maintains both the Java and C clients, to hopefully shed some light why some clients have a separate type and others don't.

jhecking:
Here's what Brian had to say on the topic of the HLLValue type:

> The java client does not include the particle type when returning values to the user.  If HLLValue did not exist, a byte[] would have been returned instead.  The problem is when the user tries to use the byte[] as input to another command.  Value.get(Object value) would not know that the byte[] is really an HLL particle type.  HLLValue's real purpose is to identify the bytes as HLL bytes.  GeoJSONValue exists for the same reason to distinguish GeoJSON strings from regular strings.
>
> The C client returns as_bytes which includes the particle type, so HLLValue is not needed.  If I were able to redesign the java client today, I would return the particle type with each value (or just return Value objects instead of raw objects).

For the Rust client, that leaves us with two options: We could either extend the existing `Value::Blob` type to include the particle type, e.g. `ParticleType::BLOB`, `ParticleType::HLL`, etc. Or we could add a separate `Value::HLL` type. I think I would opt for the latter, as it's more explicit.

jhecking:
> Context is set in the Operation.

Actually, why _is_ the CDT context part of the `Operation` struct and not part of the `CdtOperation` struct? It's the `CdtOperation::estimate_size` and `CdtOperation::write_to` functions that actually handle the context.

> Policy is set in the CdtOperation struct. In addition, policy is built into the args array together with other parameters that can not be identified once inside. It also looks like policy has a required fixed position in this array. Changing policy with the chain function would cause rebuilding this whole array.

The "problem" here is that the client's data structures too closely model after the wire protocol. For a more ergonomic client API it would be better if the `Operation`/`CdtOperation` structs were modelled at a slightly higher abstraction level, and then converted to wire format at the last moment, when serializing to the buffer. But I'm not sure what the performance impact would be, if any. Also, that would off course be a pretty major rewrite of the entire operations code at this point. :-) Not sure that's worth it. Esp. since close alignment between the different Aerospike client APIs also has some value.

jonas32:
> Actually, why is the CDT context part of the Operation struct and not part of the CdtOperation struct? It's the CdtOperation::estimate_size and CdtOperation::write_to functions that actually handle the context.

It has two reasons. First, i was not sure if it only works with CDT Operations or also with for example the scalar operations on lists. 
Second, if i would put it in the CdtOperation itself, OperationData and the CdtOperation itself would always have to be mutable. The set_context function would still have to stay in the mod.rs, because its required for chaining on operation functions.
Otherwise you would have to do it on the OperationData by matching for the right CdtType. Thats why i set it on the Operation and pass it to the CdtOperation when building the buffers.


>The "problem" here is that the client's data structures too closely model after the wire protocol. For a more ergonomic client API it would be better if the Operation/CdtOperation structs were modelled at a slightly higher abstraction level, and then converted to wire format at the last moment, when serializing to the buffer. But I'm not sure what the performance impact would be, if any. Also, that would off course be a pretty major rewrite of the entire operations code at this point. :-) Not sure that's worth it. Esp. since close alignment between the different Aerospike client APIs also has some value.

Thats i guess what this caching mechanism does in the Golang client. It builds the buffer at first execution and then saves it to the Operation for later use. I think rewriting the operations to match the other libs would be nice, but definitely a full breaking change. When we put that much effort into the operations to improve them, we should also improve the parts that could break stuff. (Policies, Contexts, Names of functions etc.)

If a full breaking change is fine, i can give it a try. Im not sure how time consuming that will be, so it might take a while.
When refactoring that much code, we should probably also talk about what else needs to be done for a 1.0 release.

jonas32:
>For the Rust client, that leaves us with two options: We could either extend the existing Value::Blob type to include the particle type, e.g. ParticleType::BLOB, ParticleType::HLL, etc. Or we could add a separate Value::HLL type. I think I would opt for the latter, as it's more explicit.

I think it would probably be easier and more straightforward to do it with `Value::HLL`, but i never used it so im ok with both options.

src/operations/lists.rs
```diff
@@ -31,65 +31,288 @@
...
+    ReverseRank = 4,
+    /// Return count of items selected.
+    Count = 5,
+    /// Return value for single key read and value list for range read.
+    Values = 7,
+    /// Invert meaning of list command and return values.
+    /// With the INVERTED flag enabled, the items outside of the specified index range will be returned.
+    /// The meaning of the list command can also be inverted.
+    /// With the INVERTED flag enabled, the items outside of the specified index range will be removed and returned.
+    Inverted = 0x10000,
```
jhecking, line 119: The `Inverted` "return type" is another good example, how the format of the wire format bleeds through into the API design, leading to a client API that's much more difficult for developers to understand/use.

E.g. if I have a bin with a list of tags (e.g. "red", "green", "blue") and I want to remove all tags _but_ the "green" tag, I can use the `remove_by_value` operation. But I have to specify a `ListReturnType.None` with the `ListReturnType.Inverted` flag set to do so, which doesn't really make much sense. After all, I don't really care about the return values at all, but the `Inverted` flag also affects the `remove` operation itself!

It would be much nicer, if the list operation itself was its own type, that had an `invert_selection` method. When serializing the operation to the wire, the client would still combine the return type and the `inverted` flag into the same byte value, of course.

jhecking: Example again for how this looks like in the Node.js client:

```
  await client.put(key, { tags: ['red', 'green', 'orange', 'blue'] })
  const result = await client.operate(key, [
    lists.removeByValue('tags', 'green')
      .invertSelection()
      .andReturn(returnType.VALUE)
  ])
  console.log(result.bins) // => { tags: [ 'red', 'orange', 'blue' ] }
  const record = await client.get(key)
  console.log(record.bins) // => { tags: [ 'green' ] }
```

The downside, of course, is that there is additional over-head for the extra function calls to construct the operation, as well as potentially the additional memory required to keep the return type and the inverted flag in two separate fields of the list operation struct, instead of combining them.

I think it's a worthwhile trade-off for a "nicer" client API. But would be best to run some micro-benchmarks to get a better sense of the actual performance overhead.

src/operations/mod.rs
```diff
@@ -41,18 +44,22 @@ pub enum OperationType {
     Append = 9,
     Prepend = 10,
     Touch = 11,
+    BitRead = 12,
+    BitWrite = 13,
+    Delete = 14,
+    HllRead = 15,
+    HllWrite = 16,
```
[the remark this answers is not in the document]
jonas32: There is only one match that takes in the OperationType. Its the buffer.rs line 402.
But there is no placeholder for HLL because it used a catch all in the end anyways.
As far as i remember, the HLL enum types should never be used yet and no changes have been made for HLL yet.

src/operations/lists.rs
jonas32: Ah ok, i get it. I agree it would make more sense to split that up. Now that its not the first time, the Operation <-> CdtOperation problem appeared. Maybe we should target a change of that before a 1.0 release. I would not like to do it in this PR, because this will probably become a similar large change and would be even harder to review.

jonas32:
> I think it would be better to add support for predicate expressions in a separate PR and to just keep this PR focused on map/list/bitmap operations. As such, I have not yet reviewed the changes you did to e.g. the scan policy, and other changes purely needed to support predicate expressions.

i added it here, because i found them missing while changing the encoder. I didnt want to open a whole new PR for this few lines but I can remove them here and move them to an own one.

src/operations/lists.rs
jhecking: Agree. We should proceed with this PR as is, and do any refactoring (if we decided to do so) in a separate PR.

jonas32:
PredExp is removed from Policy and Scan again.
Operations still has it, because that one fits the context of this PR.
I will not make a new PR for Scan and Policies again. Would have to be removed with the next version anyways.

jhecking:
Do you mind pulling from the [`aerospike-client-rust:update-operations`](https://github.com/aerospike/aerospike-client-rust/tree/update-operations) branch? I've fixed a few minor lint warnings and added/updated the copyright statements in all modified files. But I don't have permission to push to your `is-it-fresh:update-operations` branch directly.

Once that's done, this is ready to merge, right?

jonas32:
Yep, looks like this is done.

rbotzer:
Thank you @jonas32

Issue #82, opened 2020-08-28
Memory leaks


khachaturh:
The Client::batch_get function seems to make a lot memory leaks. After thousands call the process memory grow up to 10 GB. The same requests with simply Client::get looks good.  The target is x86_64-unknown-linux-gnu

jhecking:
I've been trying to replicate this, but was unsuccessful so far. So far, I've tried running the client's `batch_get` test in a loop, and observing memory allocations after every X iterations:

```
$ cargo test --release --test lib batch -- --nocapture -Zunstable-options --report-time
   Compiling aerospike v0.5.0 (/Users/jhecking/aerospike/aerospike-client-rust)
    Finished release [optimized] target(s) in 4.72s
     Running target/release/deps/lib-234f38fdc32b270d

running 1 test
       0 iterations:  8215696 bytes allocated / 13152256 bytes resident
   10000 iterations:  9496168 bytes allocated / 19107840 bytes resident
   20000 iterations:  9601664 bytes allocated / 18993152 bytes resident
   30000 iterations: 11312288 bytes allocated / 19124224 bytes resident
   40000 iterations: 11322856 bytes allocated / 18956288 bytes resident
   50000 iterations:  9536992 bytes allocated / 19046400 bytes resident
   60000 iterations:  9628136 bytes allocated / 19099648 bytes resident
test src::batch::batch_get ... test src::batch::batch_get has been running for over 60 seconds
   70000 iterations: 11179760 bytes allocated / 19070976 bytes resident
   80000 iterations: 11339816 bytes allocated / 19103744 bytes resident
   90000 iterations: 11374760 bytes allocated / 19075072 bytes resident
  100000 iterations:  9530040 bytes allocated / 18935808 bytes resident
  110000 iterations:  9599800 bytes allocated / 19161088 bytes resident
  120000 iterations: 10807808 bytes allocated / 19156992 bytes resident
  130000 iterations: 10472288 bytes allocated / 18882560 bytes resident
  140000 iterations:  8614328 bytes allocated / 18944000 bytes resident
  150000 iterations:  8367832 bytes allocated / 18849792 bytes resident
  160000 iterations:  9397392 bytes allocated / 18989056 bytes resident
  170000 iterations: 10029000 bytes allocated / 18964480 bytes resident
  180000 iterations: 10152560 bytes allocated / 18952192 bytes resident
  190000 iterations:  8348952 bytes allocated / 19030016 bytes resident
[...]

PR #83, opened 2020-09-06
Fix issue: Memory leaks #82

soro:
Make sure to shrink the allocation of the buffers once they have grown past a certain size and we are trying to truncate them. The value is probably something one might want to tune.

### DIFF
--- a/src/commands/buffer.rs
+++ b/src/commands/buffer.rs
@@ -126,9 +126,9 @@ impl Buffer {
             bail!("Invalid size for buffer: {}", size);
         }
 
-        let len = self.data_buffer.len();
+        let mem_size = self.data_buffer.capacity();
         self.data_buffer.resize(size, 0);
-        if len > 8192 && size < len {
+        if mem_size > 8192 && size < mem_size {
             self.data_buffer.shrink_to_fit();
         }
 
--- a/src/commands/buffer.rs
+++ b/src/commands/buffer.rs
@@ -99,13 +99,15 @@ const MAX_BUFFER_SIZE: usize = 1024 * 1024 + 8; // 1 MB + header
 pub struct Buffer {
     pub data_buffer: Vec<u8>,
     pub data_offset: usize,
+    pub reclaim_threshold: usize,
 }
 
 impl Buffer {
-    pub fn new() -> Self {
+    pub fn new(reclaim_threshold: usize) -> Self {
         Buffer {
             data_buffer: Vec::with_capacity(1024),
             data_offset: 0,
+            reclaim_threshold,
         }
     }
 
@@ -128,7 +130,7 @@ impl Buffer {
 
         let mem_size = self.data_buffer.capacity();
         self.data_buffer.resize(size, 0);
-        if mem_size > 8192 && size < mem_size {
+        if mem_size > self.reclaim_threshold && size < mem_size {
             self.data_buffer.shrink_to_fit();
         }
 
--- a/src/commands/buffer.rs
+++ b/src/commands/buffer.rs
@@ -126,7 +126,11 @@ impl Buffer {
             bail!("Invalid size for buffer: {}", size);
         }
 
+        let len = self.data_buffer.len();
         self.data_buffer.resize(size, 0);
+        if len > 8192 && size < len {
+            self.data_buffer.shrink_to_fit();
+        }
 
         Ok(())
     }
--- a/src/policy/client_policy.rs
+++ b/src/policy/client_policy.rs
@@ -99,7 +99,7 @@ impl Default for ClientPolicy {
             use_services_alternate: false,
             thread_pool_size: 128,
             cluster_name: None,
-            buffer_reclaim_threshold: 8192,
+            buffer_reclaim_threshold: 65536,
         }
     }
 }
--- a/src/policy/client_policy.rs
+++ b/src/policy/client_policy.rs
@@ -45,6 +45,13 @@ pub struct ClientPolicy {
     /// Throw exception if host connection fails during addHost().
     pub fail_if_not_connected: bool,
 
+    /// Threshold at which the buffer attached to the connection will be shrunk by deallocating
+    /// memory instead of just resetting the size of the underlying vec.
+    /// Should be set to a value that covers as large a percentile of payload sizes as possible,
+    /// while also being small enough not to occupy a significant amount of memory for the life
+    /// of the connection pool.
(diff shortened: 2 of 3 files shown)


src/commands/buffer.rs
```diff
@@ -126,7 +126,11 @@ impl Buffer {
             bail!("Invalid size for buffer: {}", size);
         }
 
+        let len = self.data_buffer.len();
```
jhecking, line 129: Shouldn't we check the capacity of the `Vec` rather than its current length?

src/commands/buffer.rs
```diff
@@ -126,7 +126,11 @@ impl Buffer {
...
         self.data_buffer.resize(size, 0);
+        if len > 8192 && size < len {
+            self.data_buffer.shrink_to_fit();
```
jhecking, line 132: I'm not sure if it's worth the effort, but might be better to use the [`Vec#shrink_to` function](https://doc.rust-lang.org/1.45.2/std/vec/struct.Vec.html#method.shrink_to) _if_ it is available. It's currently only available on nightly and it's not clear when/if it will get stabilized.

jhecking (commented):
Thanks for finding the issue and submitting a PR to fix it!

I'm mostly ok with the proposed fix. Just two questions

src/commands/buffer.rs
soro: Yeah, that would be better. I'll change it.

src/commands/buffer.rs
soro: Yeah, shrink_to would indeed be better, but I think this is fine for now and I wanted to make sure that it would be fixed for anyone who runs on stable as well.

soro: Might be worth putting a //TODO in there, dunno.

jhecking: Agree that the priority should be to fix the issue on stable. Not really sure how you would use `shrink_to` or `shrink_to_fit` conditionally based on whether the feature is available or not? Think you're probably right and we should just add a TODO to optimize this later if/when the feature becomes available on stable rust.

src/commands/buffer.rs
```diff
@@ -126,7 +126,11 @@ impl Buffer {
             bail!("Invalid size for buffer: {}", size);
         }
 
+        let mem_size = self.data_buffer.capacity();
         self.data_buffer.resize(size, 0);
+        if mem_size > 8192 && size < mem_size {
```
jhecking, line 131: One other thought: I think we should add another threshold to only shrink the buffer if we can reclaim at least a certain amount of memory by doing so. Reallocating the buffer is a relatively expensive operation, and it's not worth doing to save just a few bytes.

We should also uses consts to define these limits:

```
const BUFFER_RECLAIM_THRESHOLD: usize = 8192; // reclaim memory if buffer exceeds this size
const BUFFER_MIN_RECLAIM_BYTES: usize = 1024; // only reclaim if we can free at least this much memory
```

```suggestion
        if mem_size > BUFFER_RECLAIM_THRESHOLD && mem_size - size > BUFFER_MIN_RECLAIM_BYTES {
```

soro: I'll make it a constant. I kind of assumed you'd change the code with your preferred naming etc anyway, so I didn't bother. I don't mind doing it though, so I'll update later.

I am not really sold on the reclaim threshold, because at the end of the day this has the same problem. What if you can reclaim threshold+8bytes. Might as well have set the threshold higher in the first place. I think there this a good reason to make it a config setting on the client. You should know what the typical 99%ile of your payload size is and be able to size buffers accordingly. I can see if I can make that change too if you'd prefer.

soro: Looking at the code I could make it property of both the Buffer as well as ClientPolicy. That way it would be configurable and it would require relatively few changes.

khaf: IMO we should make these settings public and dynamic so that users can tweak them, and then set intelligent defaults for them. Making them `const` will guarantee that the client will perform significantly worse for some users.

jhecking: I agree that `BUFFER_RECLAIM_THRESHOLD` should be configurable and the ClientPolicy is the right place for that.

Also, if you have a fairly predictable, and mostly constant payload size, then `BUFFER_RECLAIM_THRESHOLD` is indeed sufficient, because you can just set it slightly higher than your expected max. payload size. I think the `BUFFER_MIN_RECLAIM_BYTES` would be useful if that's not the case, i.e. if your payload size varies a lot. In that case the min. reclaim bytes limit would reduce the number of buffer reallocations the client has to do. It's a trade-off between memory usage and performance.

soro:
Pushed the change. I am not convinced it is worth tuning the default parameter too much or spending too much effort on coming up with a smart way of picking that value based on statistics when really that should be something set by the user of the library according to their needs. Any automatic mechanism is also more prone to error in extreme scenarios. A sufficiently high default should be good enough in my opinion. Realistically we could probably make that buffer a max of say 100k, since that is unlikely to every lead to problems with reasonable connection pool sizes.

jhecking:
I think Khosrow just meant setting a sensible default value and not trying to do anything more complicated like trying dynamically adjust the value based on statistics. These changes now look good and ready to merge to me.

jhecking (approved):
👍

src/policy/client_policy.rs
```diff
@@ -92,6 +99,7 @@ impl Default for ClientPolicy {
             use_services_alternate: false,
             thread_pool_size: 128,
             cluster_name: None,
+            buffer_reclaim_threshold: 8192,
```
khaf, line 102: IMO this value is too small and will lead to memory thrashing for many users. In the Go client, the default is 64KiB. For a 16 node cluster with maximum number of default connections (256), it amounts to `16 x 256 x 64Kib = 256 MiB`, which is not an unreasonable amount. Keep in mind that Aerospike is a giant hash table in the end, and people end up putting a lot of data inside every record due to lack of normalization.

The most memory efficient way of keeping buffers without potential thrashing would be a multi-sized pool with idle timeouts, but that would be a bit more complicated and also could lead to potential lock contentions.

If you are concerned with the 64Kib size, 32KiB would also work.

khaf (approved):

jhecking:
Thanks again, @soro!

The Travis CI tests with Rust 1.38 are failing, but that's due to an updated `subtle` dependency, which requires Rust 1.41 in the latest version.

PR #93, opened 2021-01-25
Fix for Expression Test

jonas32:
- Adjusted the Device Size test to also support not in-memory DBs. ( #91 )
- Fixed returns for HLL from the new Server version (5.4)

### DIFF
--- a/tests/src/exp.rs
+++ b/tests/src/exp.rs
@@ -240,7 +240,16 @@ fn expression_rec_ops() {
         le(device_size(), int_val(0)),
         &set_name,
     );
-    let count = count_results(rs);
+    let mut count = count_results(rs);
+
+    if count == 0 {
+        // Not in-memory
+        let rs = test_filter(
+            le(device_size(), int_val(2000)),
+            &set_name,
+        );
+        count = count_results(rs);
+    }
     assert_eq!(count, 100, "DEVICE SIZE Test Failed");
 
     let rs = test_filter(
--- a/tests/src/exp.rs
+++ b/tests/src/exp.rs
@@ -260,7 +260,7 @@ fn expression_rec_ops() {
     assert_eq!(count, 100, "LAST UPDATE Test Failed");
 
     let rs = test_filter(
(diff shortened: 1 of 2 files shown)


jonas32:
The failing test is caused by Aerospike being too fast...
I will just set the last update ms count down to at least 10ms age instead of 150ms.
If anyone has a better idea to do that check, please tell me.

kportertx:
Why not drop the check down to 1 ms?

jonas32:
I wanted to leave at least some time for it to not get false positives. Getting something around 10ms already looks shady to me.

jhecking:
> By the way, i saw that circle is testing against ubuntu xenial. Is there any specific reason for that or is it just because no one updated it?

The CI env was last updated from Trusty to Xenial in Nov 2019. I just checked, and Xenial is still the default distro for Travis CI even now: https://docs.travis-ci.com/user/reference/linux/. That said, I don't see a reason not to switch to Bionic or even Focal. But what would be the benefit? (And assuming that the client still works and the tests still pass on those newer platforms, of course.)

jhecking:
> Fixed returns for HLL from the new Server version (5.4)

@kportertx Any idea why the number of records returned would have changed between 5.4 and earlier server versions?

kportertx:
Yes, this is related to the rust client's msgpack not packing efficiently bug. The HLL library assumed efficient packing was enforced by the server's msgpack library, because it wasn't the hashed values for rust could be different from other clients. 5.4 corrected this by detecting when values were not packed efficiently and repacking them before hashing. There wasn't a way to fix this problem that would be bug compatible cross clients.

Rust is the only client that I'm aware of that was inefficient for several types. Other clients had an inefficiency when packing bin/strings > 32 bytes and < 255 bytes which were being packed as str16 instead of str8.

jonas32:
I will create a PR to update some rust stuff anyways (mainly the clippy warnings for Rust 1.49).
When i do that, ill have a look on the msgpack bug.

jhecking:
This looks good to me. Ready to merge @jonas32?

PR #94, opened 2021-01-27
Update for rust 1.49 and fix for integer msgpack bug

### LINKED ISSUE #92: Message pack appears to calculate integer boundaries incorrectly
There are several instances of this specific pattern, though I only find it in the one file.  

See:  https://github.com/aerospike/aerospike-client-rust/blob/master/src/msgpack/encoder.rs#L257

And a playground:  https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b2ebb5a04b0a8464862963e413f57155

In this line, it kind of looks like `2 ^ 16` implies 2 raised to the power of 16.  However, Rust will calculate this as a bitwise operation and will result in 18.  So I think in this specific instance, if a value is less than 16, then we'll pack a half byte.  If it's between 16 and 18, then we'll pack an i16.  Otherwise, we'll pack an i32.  And my assumption here is that we're nearly always packing an i32.  This isn't a "breaking" bug, but it definitely would lead to inefficiencies for some use-cases and it should be really easy to build a test-case.

I looked at git blame, and I think this has been a bug for several years.

jonas32:
- Fixed all clippy warnings for version 1.49
- Resolves #92

### DIFF
--- a/src/msgpack/encoder.rs
+++ b/src/msgpack/encoder.rs
@@ -206,26 +206,26 @@ const MSGPACK_MARKER_NI64: u8 = 0xd3;
 // This method is not compatible with MsgPack specs and is only used by aerospike client<->server
 // for wire transfer only
 #[doc(hidden)]
-pub fn pack_raw_u16(buf: &mut Option<&mut Buffer>, val: u16) -> Result<usize> {
+pub fn pack_raw_u16(buf: &mut Option<&mut Buffer>, value: u16) -> Result<usize> {
     if let Some(ref mut buf) = *buf {
-        buf.write_u16(val)?;
+        buf.write_u16(value)?;
     }
     Ok(2)
 }
 
 #[doc(hidden)]
-pub fn pack_half_byte(buf: &mut Option<&mut Buffer>, val: u8) -> Result<usize> {
+pub fn pack_half_byte(buf: &mut Option<&mut Buffer>, value: u8) -> Result<usize> {
     if let Some(ref mut buf) = *buf {
-        buf.write_u8(val)?;
+        buf.write_u8(value)?;
     }
     Ok(1)
 }
 
 #[doc(hidden)]
-pub fn pack_byte(buf: &mut Option<&mut Buffer>, marker: u8, val: u8) -> Result<usize> {
+pub fn pack_byte(buf: &mut Option<&mut Buffer>, marker: u8, value: u8) -> Result<usize> {
     if let Some(ref mut buf) = *buf {
         buf.write_u8(marker)?;
-        buf.write_u8(val)?;
+        buf.write_u8(value)?;
     }
     Ok(2)
 }
@@ -239,9 +239,9 @@ pub fn pack_nil(buf: &mut Option<&mut Buffer>) -> Result<usize> {
 }
 
 #[doc(hidden)]
-pub fn pack_bool(buf: &mut Option<&mut Buffer>, val: bool) -> Result<usize> {
+pub fn pack_bool(buf: &mut Option<&mut Buffer>, value: bool) -> Result<usize> {
     if let Some(ref mut buf) = *buf {
-        if val {
+        if value {
             buf.write_u8(MSGPACK_MARKER_BOOL_TRUE)?;
         } else {
             buf.write_u8(MSGPACK_MARKER_BOOL_FALSE)?;
@@ -254,7 +254,7 @@ pub fn pack_bool(buf: &mut Option<&mut Buffer>, val: bool) -> Result<usize> {
 pub fn pack_map_begin(buf: &mut Option<&mut Buffer>, length: usize) -> Result<usize> {
     match length {
         val if val < 16 => pack_half_byte(buf, 0x80 | (length as u8)),
-        val if val >= 16 && val < 2 ^ 16 => pack_i16(buf, 0xde, length as i16),
+        val if (16..1 << 16).contains(&val) => pack_i16(buf, 0xde, length as i16),
         _ => pack_i32(buf, 0xdf, length as i32),
     }
 }
@@ -263,7 +263,7 @@ pub fn pack_map_begin(buf: &mut Option<&mut Buffer>, length: usize) -> Result<us
 pub fn pack_array_begin(buf: &mut Option<&mut Buffer>, length: usize) -> Result<usize> {
     match length {
         val if val < 16 => pack_half_byte(buf, 0x90 | (length as u8)),
-        val if val >= 16 && val < 2 ^ 16 => pack_i16(buf, 0xdc, length as i16),
+        val if (16..1 << 16).contains(&val) => pack_i16(buf, 0xdc, length as i16),
         _ => pack_i32(buf, 0xdd, length as i32),
     }
 }
@@ -272,57 +272,57 @@ pub fn pack_array_begin(buf: &mut Option<&mut Buffer>, length: usize) -> Result<
 pub fn pack_byte_array_begin(buf: &mut Option<&mut Buffer>, length: usize) -> Result<usize> {
     match length {
         val if val < 32 => pack_half_byte(buf, 0xa0 | (length as u8)),
-        val if val >= 32 && val < 2 ^ 16 => pack_i16(buf, 0xda, length as i16),
+        val if (32..1 << 16).contains(&val) => pack_i16(buf, 0xda, length as i16),
         _ => pack_i32(buf, 0xdb, length as i32),
     }
 }
 
 #[doc(hidden)]
-pub fn pack_blob(buf: &mut Option<&mut Buffer>, val: &[u8]) -> Result<usize> {
-    let mut size = val.len() + 1;
+pub fn pack_blob(buf: &mut Option<&mut Buffer>, value: &[u8]) -> Result<usize> {
+    let mut size = value.len() + 1;
 
     size += pack_byte_array_begin(buf, size)?;
     if let Some(ref mut buf) = *buf {
         buf.write_u8(ParticleType::BLOB as u8)?;
-        buf.write_bytes(val)?;
+        buf.write_bytes(value)?;
     }
 
     Ok(size)
 }
 
 #[doc(hidden)]
-pub fn pack_string(buf: &mut Option<&mut Buffer>, val: &str) -> Result<usize> {
-    let mut size = val.len() + 1;
+pub fn pack_string(buf: &mut Option<&mut Buffer>, value: &str) -> Result<usize> {
+    let mut size = value.len() + 1;
 
     size += pack_byte_array_begin(buf, size)?;
     if let Some(ref mut buf) = *buf {
         buf.write_u8(ParticleType::STRING as u8)?;
-        buf.write_str(val)?;
+        buf.write_str(value)?;
     }
 
     Ok(size)
 }
 
 #[doc(hidden)]
-pub fn pack_raw_string(buf: &mut Option<&mut Buffer>, val: &str) -> Result<usize> {
-    let mut size = val.len();
+pub fn pack_raw_string(buf: &mut Option<&mut Buffer>, value: &str) -> Result<usize> {
+    let mut size = value.len();
 
     size += pack_byte_array_begin(buf, size)?;
     if let Some(ref mut buf) = *buf {
-        buf.write_str(val)?;
+        buf.write_str(value)?;
     }
 
     Ok(size)
 }
 
 #[doc(hidden)]
-fn pack_geo_json(buf: &mut Option<&mut Buffer>, val: &str) -> Result<usize> {
-    let mut size = val.len() + 1;
+fn pack_geo_json(buf: &mut Option<&mut Buffer>, value: &str) -> Result<usize> {
+    let mut size = value.len() + 1;
 
     size += pack_byte_array_begin(buf, size)?;
     if let Some(ref mut buf) = *buf {
         buf.write_u8(ParticleType::GEOJSON as u8)?;
-        buf.write_str(val)?;
+        buf.write_str(value)?;
     }
 
     Ok(size)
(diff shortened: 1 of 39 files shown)


src/msgpack/encoder.rs
```diff
@@ -254,7 +254,7 @@ pub fn pack_bool(buf: &mut Option<&mut Buffer>, val: bool) -> Result<usize> {
 pub fn pack_map_begin(buf: &mut Option<&mut Buffer>, length: usize) -> Result<usize> {
     match length {
         val if val < 16 => pack_half_byte(buf, 0x80 | (length as u8)),
-        val if val >= 16 && val < 2 ^ 16 => pack_i16(buf, 0xde, length as i16),
+        val if (16..1 << 16).contains(&val) => pack_i16(buf, 0xde, length as i16),
```
jhecking, line 257: Adding an extra set of braces might make this a bit easier to read; the `16..1` part looks quite odd at first, otherwise:

```suggestion
        val if (16..(1 << 16)).contains(&val) => pack_i16(buf, 0xde, length as i16),
```

Though I'm not sure if rustfmt would just remove the unnecessary braces again anyway? (Or maybe clippy complains?)

src/policy/write_policy.rs
```diff
@@ -71,10 +71,18 @@ pub struct WritePolicy {
...
+            record_exists_action: RecordExistsAction::Update,
+            generation_policy: GenerationPolicy::None,
+            commit_level: CommitLevel::CommitAll,
+            generation: gen,
+            expiration: exp,
+            send_key: false,
+            respond_per_each_op: false,
+            durable_delete: false,
+            filter_expression: None,
+        }
```
jhecking, line 85: I'm not quite sure I agree with this change. Was this suggested by Clippy? Now you have to maintain the default values for properties like `record_exists_action`, `generation_policy`, etc. in two separate places, rather than having them contained only in `impl Default for WritePolic`.

jhecking (commented):
Looks great! Thanks for doing this cleanup. 🙏 

Since you are introducing the standard `matches!` macro – thereby increasing the min. Rust version to  1.42 – can you please replace this custom `matches_override` macro as well: https://github.com/aerospike/aerospike-client-rust/blob/cb1d022097947fd563ad6a1ae9cd4944275ebf60/tests/src/task.rs#L25-L32

src/policy/write_policy.rs
jonas32: Yes, that was the reason i changed it. I also dont think that this makes much more sense than how it was before.

jonas32:
> Since you are introducing the standard matches! macro – thereby increasing the min. Rust version to 1.42 – can you please replace this custom matches_override macro as well

Thats this PR: #74

jhecking:
> Thats this PR: #74
> The Build is failed because the rust version used on travis was too old when the CI was executed.

At the time, the build matrix for Travis explicitly included Rust 1.38, in an attempt to maintain that as the minimum required Rust version for the client: [.travis.yml](https://github.com/aerospike/aerospike-client-rust/blob/a6b48fc67bbcd67739d22b52ad31c2085911d7c7/.travis.yml). But I gave up on that at some point last year, because the builds kept breaking when some dependencies (often indirect ones) required newer Rust versions: https://github.com/aerospike/aerospike-client-rust/pull/84.

It you rebase https://github.com/aerospike/aerospike-client-rust/pull/74 on master branch now, it should pass.

src/policy/write_policy.rs
jhecking: Why not follow the [suggested remediation](https://rust-lang.github.io/rust-clippy/master/index.html#field_reassign_with_default) for the `field_reassign_with_default` lint and use the [functional update syntax](https://doc.rust-lang.org/reference/expressions/struct-expr.html#functional-update-syntax), e.g.

```suggestion
        Self {
            generation: gen,
            expiration: exp,
            ..WritePolicy::default()
        }
```

I'm not sure when this syntax was introduced? Maybe this would raise the min. required Rust version even more, if it was only added after Rust 1.42. But looks nice and clean, and is as maintainable as the orig.!

jonas32: I think we can go with rust 1.49 as min compiler version. As far as i know there are no breaking backwards changes.

jhecking:
I'm not quite comfortable with requiring all users to be on the latest stable Rust version, which was only just released less than a month ago. I'm still not sure when the "functional update syntax" for structs was introduced, but some of the `const fn` changes you are introducing in this PR require Rust 1.46.0 (released Aug'20). That's the oldest version I was able to compile the client with on this branch.

For comparison, the Aerospike Node.js client still runs fine on Node.js 10, which was first released in 2018. That might not be a fair comparison, since each language/community handles long-term support and backward compatibility very differently. My initial goal for the Rust client was to support stable versions going back at least a year, but even that has been a challenge. And I'm still not quite sure what's "expected" or "normal" within the Rust community. I guess as long as it does not become a blocker for potential users to start using the Aerospike client, it's ok.

jonas32:
Most of the bigger rust projects only say "compatible with the latest version". I have not seen anyone grant that backwards compatibility at least. 
I guess rust is probably too "young" and the development of the language is too agile to be able to give such compatibility.
There might be big changes in a future release that introduce big performance changes. If we now do that grant, we will have to always do. That might be a bigger business killer for the client than not granting it.
For example see the clippy log of the output on that Issue. As a customer i would probably have a bad feeling using a lib in my software that prints so many warnings.

In addition, we just did a first stable release two month ago. I guess the node one already had a stable release back then.
Going back that far might be a performance factor too. People who use rust for that kind of infrastructure are probably looking for the performance and not the compatibility. In that case you would probably go for Java or Node.
Const Functions might not be a big performance factor, but they can be a factor.

jonas32:
I updated the macro in this PR. I somehow managed to kill the other branch while rebasing...

src/msgpack/encoder.rs
jonas32: Done

brianbruggeman:
Are we getting rid of _all_ of the clippy errors?

```
$ git remote add is-it-fresh https://github.com/is-it-fresh/aerospike-client-rust
$ git fetch is-it-fresh
$ git checkout update/rust-1.49.0
$ cargo clippy
    Checking aerospike v1.0.0 (/Users/brianbruggeman/repos/others/aerospike-client-rust)
warning: this argument (1 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)
  --> src/expressions/bitwise.rs:64:13
   |
64 |     policy: &BitPolicy,
   |             ^^^^^^^^^^ help: consider passing by value instead: `BitPolicy`
   |
note: the lint level is defined here
  --> src/lib.rs:25:22
   |
25 | #![warn(clippy::all, clippy::pedantic, clippy::nursery)]
   |                      ^^^^^^^^^^^^^^^^
   = note: `#[warn(clippy::trivially_copy_pass_by_ref)]` implied by `#[warn(clippy::pedantic)]`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

warning: this argument (1 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)
  --> src/expressions/bitwise.rs:97:13
   |
97 |     policy: &BitPolicy,
   |             ^^^^^^^^^^ help: consider passing by value instead: `BitPolicy`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

warning: this argument (1 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)
   --> src/expressions/bitwise.rs:128:13
    |
128 |     policy: &BitPolicy,
    |             ^^^^^^^^^^ help: consider passing by value instead: `BitPolicy`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

warning: this argument (1 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)
   --> src/expressions/bitwise.rs:161:13
    |
161 |     policy: &BitPolicy,
    |             ^^^^^^^^^^ help: consider passing by value instead: `BitPolicy`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

warning: this argument (1 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)
   --> src/expressions/bitwise.rs:188:13
    |
188 |     policy: &BitPolicy,
    |             ^^^^^^^^^^ help: consider passing by value instead: `BitPolicy`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

warning: this argument (1 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)
   --> src/expressions/bitwise.rs:215:13
    |
215 |     policy: &BitPolicy,
    |             ^^^^^^^^^^ help: consider passing by value instead: `BitPolicy`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

warning: this argument (1 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)
   --> src/expressions/bitwise.rs:242:13
    |
242 |     policy: &BitPolicy,
    |             ^^^^^^^^^^ help: consider passing by value instead: `BitPolicy`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

warning: this argument (1 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)
   --> src/expressions/bitwise.rs:267:13
    |
267 |     policy: &BitPolicy,
    |             ^^^^^^^^^^ help: consider passing by value instead: `BitPolicy`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

warning: this argument (1 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)
   --> src/expressions/bitwise.rs:291:13
    |
291 |     policy: &BitPolicy,
    |             ^^^^^^^^^^ help: consider passing by value instead: `BitPolicy`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

warning: this argument (1 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)
   --> src/expressions/bitwise.rs:317:13
    |
317 |     policy: &BitPolicy,
    |             ^^^^^^^^^^ help: consider passing by value instead: `BitPolicy`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

warning: this argument (1 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)
   --> src/expressions/bitwise.rs:346:13
    |
346 |     policy: &BitPolicy,
    |             ^^^^^^^^^^ help: consider passing by value instead: `BitPolicy`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

warning: this argument (1 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)
   --> src/expressions/bitwise.rs:382:13
    |
382 |     policy: &BitPolicy,
    |             ^^^^^^^^^^ help: consider passing by value instead: `BitPolicy`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

warning: this argument (1 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)
   --> src/expressions/bitwise.rs:416:13
    |
416 |     policy: &BitPolicy,
    |             ^^^^^^^^^^ help: consider passing by value instead: `BitPolicy`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

warning: this argument (2 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)
  --> src/expressions/maps.rs:27:13
   |
27 |     policy: &MapPolicy,
   |             ^^^^^^^^^^ help: consider passing by value instead: `MapPolicy`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

warning: this argument (2 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)
  --> src/expressions/maps.rs:56:13
   |
56 |     policy: &MapPolicy,
   |             ^^^^^^^^^^ help: consider passing by value instead: `MapPolicy`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

warning: this argument (2 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)
  --> src/expressions/maps.rs:83:13
   |
83 |     policy: &MapPolicy,
   |             ^^^^^^^^^^ help: consider passing by value instead: `MapPolicy`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

warning: this argument (2 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)
   --> src/operations/maps.rs:195:42
    |
195 | pub(crate) const fn map_write_op(policy: &MapPolicy, multi: bool) -> CdtMapOpType {
    |                                          ^^^^^^^^^^ help: consider passing by value instead: `MapPolicy`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

warning: this argument (2 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)
   --> src/operations/maps.rs:221:32
    |
221 | const fn map_order_arg(policy: &MapPolicy) -> Option<CdtArgument> {
    |                                ^^^^^^^^^^ help: consider passing by value instead: `MapPolicy`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

warning: 18 warnings emitted

    Finished dev [unoptimized + debuginfo] target(s) in 3.42s
```

jonas32:
Yes i will fix them too. The warnings you see are new because of the last commit adding Clone.

jhecking:
> Yes i will fix them too. The warnings you see are new because of the last commit adding Clone.
> This would be a breaking change... @jhecking what do you say?

I think we should continue to pass MapPolicy and BitPolicy as reference for consistency. We do pass all other policies as references as well, e.g.

    pub fn get<T>(&self, policy: &ReadPolicy, key: &Key, bins: T) -> Result<Record>

Those other policies probably exceed the 8 byte limit, below which clippy's [`trivially_copy_pass_by_ref`](https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref) lint triggers. I would suggest to disable that lint for these two policies as well.

Unless you see an actual benefit to this change, and not just to satisfy clippy?

jonas32:
The reason would only be clippy. Then i will just add ignores for that.

jhecking:
Thanks, @jonas32! Are you planning any further changes or is this ready to be merged?

jhecking (approved):

jonas32:
If you have no changes left, please merge.

Issue #95, opened 2021-01-30
Normalizing and standardizing environments


brianbruggeman:
Given the [issues I had experienced with just the testing aspect](https://github.com/aerospike/aerospike-client-rust/issues/91), I am proposing here a Makefile shim which standardizes specific, common development operations including but not limited to:

* builds
* qa
* style
* testing

The primary goal here is to limit the amount of ramp-up time anyone new would need to have in order to start contributing and simultaneously reduce the differences across development environments so that debugging differences becomes simple.  

Additionally, I think the shim could also be used in CI though I'm not sure I have actually access to the Travis output (it looked like notifications were sent to a team specifically at Aerospike).  Despite being sanctioned under the Aerospike umbrella, since this particular project community driven, it makes me wonder if Github's Actions are a better option.  The only downside I see here is that a Makefile approach is probably limited to a Posix environment.   As a consequence, it probably wouldn't work for Appveyor/Windows.

jonas32:
Great Idea.
I dont really see a problem as long as the compatibility to running the client itself on windows stays like it is.
Right now, Appveyor is not running DB tests. Its just checking if something breaks in the compile step.
We could let the Linux tests run that Makefile and leave appveyor as it is. (I dont think anyone is going to use this client on a windows machine in production anyways).
You can at least access the direct output of the builds by pressing them in your pull request. I always used Draft PRs for that.
That way, the CI will also run on your branch and execute the tests.
For example here: https://travis-ci.org/github/aerospike/aerospike-client-rust/builds/756731191

jhecking:
The Travis build logs should definitely be open and are linked from the PRs, as Jonas pointed out.

I think using the Makefile in the CI builds would be a good idea - at least for the builds in Travis. However, the Travis builds currently start a 2-node Aerospike cluster, which your Makefile currently does not support. There also needs to be a mechanism to wait for the cluster to have formed, before the tests are run. Both of these things are accomplished by the [.travis/start_cluster.sh](https://github.com/aerospike/aerospike-client-rust/blob/master/.travis/start_cluster.sh) script. If you can integrate that logic into the Makefile, then I don't see a problem switching over.

Also, there's already some internal discussion about moving from Travis to GitHub Actions. But no final decision yet.

PR #96, opened 2021-01-30
Adds a functional Makefile

brianbruggeman:
Adds a simple Makefile to standardize development

Addresses #95.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,105 @@
+# ======================================================================
+# 
+#  Standardized commands 
+# 
+# ======================================================================
+.DEFAULT_GOAL := help
+
+aerospike: $(DOCKER)  ## Start aerospike docker instance
+	mkdir -p $(AEROSPIKE_DATA_PATH)
+	$(DOCKER) run --rm -it $(AEROSPIKE_DOCKER_RUN) --name aerospike $(AEROSPIKE_URL)
+
+audit: $(CARGO) $(AUDIT)  ## Check for security vulnerabilities
+	$(CARGO) audit
+
+benchmark: $(CARGO)  ## Benchmarks code
+	$(CARGO) bench
+
+build: $(CARGO)  ## Compiles crate workspace
+	$(CARGO) build --release --workspace
+
+doc: $(CARGO)  ## Build documentation
+	$(CARGO) doc --no-deps
+
+help: ## Show this menu 
+	@awk 'BEGIN {FS = ":.*?## "} /^[a-zA-Z_-]+:.*?## / {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST)
+
+lint: $(CARGO) clippy  ## Run style check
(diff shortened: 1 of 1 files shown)


jonas32:
Looks fine.
I just found one little point i would try to avoid.
You hardcoded the ports into the config. 
What do you think about making them as env vars with a default value?

brianbruggeman:
I had actually thought about adding a port variable initially, but then rejected it because:

1.  I think it would be good to test against 3 ports rather than 1.
2.  If I wanted to add three ports, I think I'd need 3 port environment variables
3.  I think AEROSPIKE_HOSTS would probably need to reflect the choice of port environment variables
4. The ports would really only be for the local system because the container would always use the default of 300x

At that point, I wondered if there was really much sense in adding a port variable.

That said I do want to make another update to the Makefile, so if you think there's enough of a good reason despite the above 4, I'm not opposed to adding it.

jonas32:
No you are right. I didnt think about just overwriting the whole variable. Thats perfectly fine.

Makefile
```diff
@@ -0,0 +1,120 @@
...
+
+# aerospike
+AEROSPIKE_DATA_PATH ?= $(HOME)/.aerospike/data
+AEROSPIKE_DEFAULT_TTL ?= 0
+AEROSPIKE_DOCKER_REPO ?= aerospike/aerospike-server
+AEROSPIKE_MEMORY ?= 1
+AEROSPIKE_NAMESPACE ?= test
+AEROSPIKE_STORAGE ?= 8
+AEROSPIKE_URL ?= $(AEROSPIKE_DOCKER_REPO):$(AEROSPIKE_VERSION)
+AEROSPIKE_VERSION ?= 5.4.0.2
```
jhecking, line 73: I think `latest` would be a better default, rather than hard-coding a specific version number.

jhecking (commented):
Thanks for your contribution! I think adding a Makefile with standarized commands to build, test, release the client is a great idea.

Makefile
brianbruggeman: @jhecking

I would like to also use `latest`, because I think that's a set it and forget it type of thing.

How do we handle the problem where `latest` has pulled from dockerhub months ago and is pointing to a (possibly) much older version?  I actually had this issue last week because on my localhost, the `latest` wasn't pointing to version 5.4 version and I don't think docker actually checks datetime timestamps and/or image hashes to see if there's an update to the `latest` tag.

I think for CI/CD, `latest` would work just fine, though.

brianbruggeman:
@jonas32 

I took your thoughts and made several port environment variables using: https://www.aerospike.com/docs/operations/configure/network/

At this point you can now override where they are resolved locally.

Makefile
jhecking: You could add `$(DOCKER) pull $(AEROSPIKE_URL)` to the `aerospike` target to ensure you are always running the latest version.

jhecking (approved):

PR #100, opened 2021-02-25
Aerospike 5.6 Expressions

jonas32:
Content:

- [x] New Basic Expressions
- [ ] New CDT Expressions
- [ ] Operation Read/Write Expressions

Requires at least Aerospike 5.6.
Dont merge before 5.6 release.

### DIFF
--- a/src/expressions/mod.rs
+++ b/src/expressions/mod.rs
@@ -68,6 +68,31 @@ pub enum ExpOp {
     And = 16,
     Or = 17,
     Not = 18,
+    Xor = 19,
+    Add = 20,
+    Sub = 21,
+    Mul = 22,
+    Div = 23,
+    Pow = 24,
+    Log = 25,
+    Mod = 26,
+    Abs = 27,
+    Floor = 28,
+    Ceil = 29,
+    ToInt = 30,
+    ToFloat = 31,
+    IntAnd = 32,
+    IntOr = 33,
+    IntXor = 34,
+    IntNot = 35,
+    IntLshift = 36,
+    IntRshift = 37,
+    IntARshift = 38,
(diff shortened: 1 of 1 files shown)


src/operations/exp.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// If bin exists, the operation will be denied.
+    /// If bin exists, fail with Bin Exists
+    CreateOnly = 1,
+    /// If bin exists, the bin will be overwritten.
+    /// If bin does not exist, the operation will be denied.
+    /// If bin does not exist, fail with Bin Not Found
+    UpdateOnly = 2,
+    /// If expression results in nil value, then delete the bin.
+    /// Otherwise, return OP Not Applicable when NoFail is not set
+    AllowDelete = 3,
```
kportertx, line 40: This should be 4. Because I am also prone to this particular mistake, I prefer to write flags this way (YMMV):

```rust
// Assuming this also works in rust.
Default = 0,
CreateOnly = 1 << 0,
UpdateOnly = 1 << 1,
AllowDelete = 1 << 2,
PolicyNoFail = 1 << 3,
EvalNoFail = 1 << 4,
```

kportertx (changes requested):

src/operations/exp.rs
jonas32: Thanks for pointing out! Fixed.

jonas32:
Tests are failing because of the Aerospike Version. Tested against 5.5.0.2-111.
Some minor changes (comments, formatting) are still pending, but i think its review ready.

khaf:
@jonas32 Is there a reason this branch is not merged yet?

PR #116, opened 2022-02-24
Fixed large integers packing when encoding to messagepack

### LINKED ISSUE #115: `Parameter error` when filtering on large i64 values
Hello, I've been fighting with this issue for some time now, and cannot seem to understand what's the root of the problem.
Let's imagine we have the following code:
```rust
use aerospike::{
    expressions::{eq, int_bin, int_val},
    Bin, Bins, Client, ClientPolicy, Key, QueryPolicy, Statement, Value, WritePolicy,
};

fn main() {
    let c = Client::new(&ClientPolicy::default(), &"172.17.0.1:3000").unwrap();
    let pkey = Pubkey::new("Feature111111111111111111111111111111111111");
    let acc = Account {
        owner: pkey.clone(),
        data: vec![1; 32],
    };

    let key = Key::new("test", "set", Value::Int(1)).unwrap();
    c.put(&WritePolicy::default(), &key, &acc.into_bins())
        .unwrap();
    let mut qpolicy = QueryPolicy::new();
    let filter = eq(int_bin("oidx".into()), int_val(pkey.hash() as i64));
    qpolicy.filter_expression.replace(filter);
    let stmt = Statement::new("test", "set", Bins::All);
    let set = c.query(&qpolicy, stmt).unwrap();
    for s in &*set {
        if let Err(e) = s {
            eprintln!("Error querying aerospike: {}", e);
        }
    }
}

struct Account {
    owner: Pubkey,
    data: Vec<u8>,
}

impl Account {
    fn into_bins(&self) -> [Bin<'static>; 3] {
        let owner = Bin::new("owner", Value::from(&self.owner));
        let data = Bin::new("data", Value::from(&self.data));
        let oidx = Bin::new("oidx", Value::from(self.owner.hash() as i64));

        [owner, data, oidx]
    }
}

#[derive(Clone)]
struct Pubkey(Vec<u8>);

impl From<&Pubkey> for Value {
    fn from(key: &Pubkey) -> Self {
        key.0.clone().into()
    }
}

impl Pubkey {
    fn hash(&self) -> u64 {
        use std::collections::hash_map::DefaultHasher;
        use std::hash::{Hash, Hasher};
        let mut hasher = DefaultHasher::new();
        self.0.hash(&mut hasher);
        hasher.finish()
    }

    fn new(s: &str) -> Self {
        let vec = bs58::decode(s).into_vec().unwrap();
        Self(vec)
    }
}
```
Nothing fancy, though a little lengthy, so the issue arises when one tries to query this set using filter `expressions:eq` filters on `oidx` bin, which stores `Value::Int`, library just reports that there was a `Parameter error` while on server one can observe a warning
```
WARNING (exp): (exp.c:1282) build_count_sz - invalid op_code 105 size 0
WARNING (scan): (scan.c:706) basic scan job failed filter processing
```
What makes this issue interesting, is that it only happens with large positive numbers, yes, with negatives it doesn't happen, no matter how large the number is (e.g. hash of `Sysvar1111111111111111111111111111111111111`), also it doesn't happen with small numbers (not sure what the range is, but it's probably pretty large). I would've thought that this might be a server issue (aerospike works this way may be), but if you make this query via aql 
```
SELECT * FROM test.set WHERE oidx = 327397455200479690
```
Then everything works out fine. Additionally if one were to use `Statement::add_filter` to add the same filter, then it works with any valid i64 values, without causing `Parameter error` issue. 

I'll try to provide any additional context if necessary, thanks.

bobs4462:
Closes #115

### DIFF
--- a/src/msgpack/encoder.rs
+++ b/src/msgpack/encoder.rs
@@ -196,7 +196,7 @@ const MSGPACK_MARKER_BOOL_FALSE: u8 = 0xc2;
 const MSGPACK_MARKER_I8: u8 = 0xcc;
 const MSGPACK_MARKER_I16: u8 = 0xcd;
 const MSGPACK_MARKER_I32: u8 = 0xce;
-// const MSGPACK_MARKER_I64: u8 = 0xd3;
+const MSGPACK_MARKER_I64: u8 = 0xd3;
 
 const MSGPACK_MARKER_NI8: u8 = 0xd0;
 const MSGPACK_MARKER_NI16: u8 = 0xd1;
@@ -341,7 +341,7 @@ pub fn pack_integer(buf: &mut Option<&mut Buffer>, val: i64) -> Result<usize> {
         val if val >= i64::from(i16::max_value()) && val < i64::from(i32::max_value()) => {
             pack_i32(buf, MSGPACK_MARKER_I32, val as i32)
         }
-        val if val >= i64::from(i32::max_value()) => pack_i64(buf, MSGPACK_MARKER_I32, val),
+        val if val >= i64::from(i32::max_value()) => pack_i64(buf, MSGPACK_MARKER_I64, val),
 
         // Negative values
         val if val >= -32 && val < 0 => {
--- a/tests/src/query.rs
+++ b/tests/src/query.rs
@@ -198,3 +198,36 @@ fn query_node() {
 
     assert_eq!(count.load(Ordering::Relaxed), 100);
 }
+
+// https://github.com/aerospike/aerospike-client-rust/issues/115
+#[test]
+fn query_large_i64() {
+    const SET: &str = "large_i64";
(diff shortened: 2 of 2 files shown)


khaf:
Thanks for your PR. Could you please include a test case to make sure the issue doesn't creep back into the code base in the future?

bobs4462:
> Thanks for your PR. Could you please include a test case to make sure the issue doesn't creep back into the code base in the future?

Done, added new test case for query with filter on large i64 value

khaf:
Could you please revert the write confirmation removal? Writes are not fire and forget and the server may return meaningful errors.

bobs4462:
Oops, I'm really sorry, it slipped through when I was experimenting on my own project, reverted it back

khaf:
Thanks for your pull request. Sorry it took us a while to release it.

PR #139, opened 2023-06-20
fix password passing

databasedav:
otherwise password must be hashed by the user, this is more ergonomic and matches the behavior of the go client


khaf:
This is a bad idea. The hashing is _very_ slow, and as you will notice in the `ClientPolicy.set_user_password()`, it is done once and cached. The Go client has two different API for authentication, and internally it uses the one that takes the hash. I think this is more of an issue with documentation, since knowing what to do is enough, and doesn't really impede the users in any other way.
I'm still open to feedback though.

databasedav:
didn't notice the `set_user_password` method, this was only an issue when manually constructing the `ClientPolicy` struct

khaf:
You didn't do anything wrong. I'll improve the documentation, and maybe I'll also remove the pub from the field so everyone has to use the method.

PR #141, opened 2023-08-31
made expOp and filterExp::new pub

vmsachin:
changed the visibility of expOp enum and new method of filterExp to public as I need to use them in the extension code

### DIFF
--- a/aerospike-core/src/query/filter.rs
+++ b/aerospike-core/src/query/filter.rs
@@ -181,7 +181,7 @@ macro_rules! as_within_radius {
         let lng = as_val!($lng as f64);
         let radius = as_val!($radius as f64);
         let geo_json = format!(
-            "{{ \"type\": \"Aeroircle\", \"coordinates\": [[{:.8}, {:.8}], {}] }}",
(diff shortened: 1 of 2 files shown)


vmsachin:
also fixed a few spelling errors

aerospike-core/src/query/filter.rs
```diff
@@ -185,7 +185,8 @@ macro_rules! as_within_radius {
             "{{ \"type\": \"AeroCircle\", \"coordinates\": [[{:.8}, {:.8}], {}] }}",
             lng, lat, radius
         );
-        let geo_string = as_val!(geo_json);
+        let geo_copy = geo_json.clone();
```
khaf, line 188: I still do not understand why this is needed.

khaf (changes requested):

aerospike-core/src/query/filter.rs
vmsachin: I have removed creating the extra variable to hold the string but the reason we need to preserve the string is because of the wire protocol. The server expects the Filter to have Value [begin, end] of type string and not GeoJSON.

PR #144, opened 2023-11-14
Authentication change - user management 

vmsachin:
1. Removed username_password field
2. Made it two separate fields
3. Added requires_auth() fn to client policy 
4. Do the entire auth routine only if requires_auth is true.


khaf (commented):
As per our call, this PR is not applicable and should be closed.

PR #146, opened 2024-02-06
Derive Clone trait for Operation struct

juliannguyen4:


khaf:
This was applied to the v2 version.

PR #168, opened 2025-11-12
CLIENT-3892: write GeoJSON in query filters as raw strings (no flags+ncells prefix)

paul-brackin:


aerospike-core/src/query/filter.rs
```diff
@@ -69,9 +69,15 @@ impl Filter {
 
     #[doc(hidden)]
     pub fn estimate_size(&self) -> usize {
-        // bin name size(1) + particle type size(1)
-        //     + begin particle size(4) + end particle size(4) = 10
-        self.bin_name.len() + self.begin.estimate_size() + self.end.estimate_size() + 10
+        let begin_size = match &self.begin {
```
khaf, line 72: @paul-brackin does this mean the Java implementation is also buggy?

khaf:
The issue was in macros.

PR #171, opened 2025-11-18
Provide Dockerfile for initial MNTN-related testing.

sfalvo-aerospike:
This Dockerfile builds a Debian Trixie base image with the latest Rust installed.  It then executes a Bash script named "startup-sequence", whose job it is to run the `cargo test` command to invoke tests.  Note that some environment variables are required when running the resulting Docker image (see Rust client docs).

As more property-based testing facilities are provided, this Dockerfile and startup-sequence may be amended to take additional configuration via environment variables.


juliannguyen4 (approved):
LGTM, this is fine since we need to get the rests running ASAP. Just something to note in the future:

- We should add a README explaining what the Dockerfile is used for
- Also you can `chmod u+x` the startup script and commit those permissions in the repo
- There's an official image for Rust: https://hub.docker.com/_/rust
- To test the Dockerfile, you can use it to run the tests in CI/CD

tools/docker/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+WORKDIR /app
+RUN apt update -y && \
+    apt install -y git curl build-essential
+
+RUN curl https://sh.rustup.rs -sSf | sh -s -- -y
+
+COPY startup-sequence /app/startup-sequence
+RUN chmod +x /app/startup-sequence
+RUN echo 'source $HOME/.cargo/env' >> $HOME/.bashrc
```
khaf: This is already sourced in the script, why add it here?

tools/docker/startup-sequence
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+source $HOME/.cargo/env
+
+git clone https://github.com/aerospike/aerospike-client-rust.git
+cd aerospike-client-rust
+git checkout tls
```
khaf: I believe the name of the branch should be an ENV variable.

tools/docker/startup-sequence
```diff
@@ -0,0 +1,11 @@
...
+cargo clean
+cargo fix --lib -p aerospike-core
```
khaf: Should be removed. Fixes should be applied to the source, otherwise tests will run on different source code.

tools/docker/startup-sequence
```diff
@@ -0,0 +1,11 @@
...
+cargo test
```
khaf: Where do the ENV variables come from? IIRC they need to be explicitly passed here, no?

khaf (changes requested):

juliannguyen4:
Also noticed that Rust client code seems to be building when running the container. Shouldn't this already be built in the image?

sfalvo-aerospike:
> Also noticed that Rust client code seems to be building when running the container. Shouldn't this already be built in the image?

This is what I was talking about yesterday regarding the `cargo test --no-run` flag.  However, I wasn't able to get it working in the Dockerfile.  So, rather than commit a broken Dockerfile, I did the "worse-is-better" thing and just let Cargo build the tests before running them.  We can always iterate and improve upon this later.

PR #174, opened 2025-12-18
Client-4015:  Allow empty set names

paul-brackin:


aerospike-core/src/query/statement.rs
```diff
@@ -127,10 +131,6 @@ impl Statement {
             }
         }
 
-        if self.set_name.is_empty() {
```
khaf, line 130: I believe removing this check and adding a comment on `Statement.set_name` is enough. No need to turn it to an optional.

khaf (changes requested):

PR #177, opened 2026-01-08
Client-4052: parse privileges misalignment

paul-brackin:
…misalignment


khaf:
What is this PR supposed to do?

PR #178, opened 2026-01-29
Client-4131: create PKI user double hashing

paul-brackin:


khaf:
Good catch, but the fix is much simpler. I'll close this one and submit the fix myself.

Here is the commit: d172739b3c22fc019aee829b4f8462fe2ba8a043

PR #182, opened 2026-02-09
Client 4113 rust performance testing as bench

dmishra-aero:
- Added different workload (RU,RR,  RMI,RMD)
- Supported multiple objects specs
- Retrofit code structure
- Fix cli option such as use-service-alternate/ ip_map that was not honoured properly
- Tokio tunning for high TPS


khaf (approved):
Looks good.

Issue #183, opened 2026-02-12
bug: v2 Record.bins must not lose bin order


KimSoungRyoul:
## Problem

current v2 client `Record.bins` is defined as `HashMap<String, Value>` in aerospike-core
(https://github.com/aerospike/aerospike-client-rust/blob/master/aerospike-core/src/record.rs#L40):
#### current impl
~~~rust
pub struct Record {
    pub key: Option<Key>,
    pub bins: HashMap<String, Value>, // <---- IndexMap<String, Value>
    pub generation: u32,
    expiration: u32,
}
~~~
HashMap does not preserve insertion order. When the Aerospike server returns bins in a specific order (e.g. in response to an operate call
 with multiple operations), that order is lost during deserialization into the Record struct.

### Comparison with other clients

| Client | `bins` storage type | Order preserved |
|--------|-------------------|-----------------|
| C | `as_bin[]` (array) | Yes |
| Java | [`LinkedHashMap<String, Object>`](https://github.com/aerospike/aerospike-client-java/blob/master/client/src/com/aerospike/client/command/Command.java) | Yes |
| Rust | `HashMap<String, Value>` ??? |  No ??? |

Both the C client (flat array) and Java client (`LinkedHashMap`) preserve server response order. The Rust client is the only one that does not.

## Environment

- aerospike-core: 2.0.0-alpha.9
- aerospike: 2.0.0-alpha.9

---

khaf:
This is a major breaking change, and not critical, since multiple operations on a bin are bunched up in a `Value::MultiResult`, so you will never confuse the returned values. But I agree that it is not a bad idea, and will implement it for the next version of the client.

PR #184, opened 2026-02-12
fix: v2 Record.bins must not lose bin order

KimSoungRyoul:
Fixes https://github.com/aerospike/aerospike-client-rust/issues/183

- Replace `HashMap<String, Value>` with `IndexMap<String, Value>` for `Record.bins` to preserve server-returned bin order


KimSoungRyoul:
> This is a major breaking change, and not critical, since multiple operations on a bin are bunched up in a `Value::MultiResult`, so you will never confuse the returned values. But I agree that it is not a bad idea, and will implement it for the next version of the client.

close pullrequest

PR #186, opened 2026-02-17
Client 4102 update readme

gagan405:
1. Update the supported rust version
2. Separate the installation sections to 2 parts: `build from source` and `use as dependency`
3. Add `rust-version` to `Cargo.toml`


khaf (approved):
Looks good.

PR #188, opened 2026-02-18
Update rust docs for client APIs

gagan405:
1. Update rust docs for client APIs
2. Fix broken examples and add examples wherever they are runnable

Validated examples through:

```
 cargo test --doc -p aerospike-core client.rs
```


khaf (approved):
Looks good.

PR #191, opened 2026-02-20
CLIENT-4156 fix rust-doc examples at remaining places in client, errors, and expressions

gagan405:
This change makes all doc-tests runnable, except for sync-client tests which need an async runtime. That will require some modifications in how we enable the features.

For now, its good to go with all compiling doctests:

```
cargo test --doc --workspace
```

### DIFF
--- a/aerospike-core/Cargo.toml
+++ b/aerospike-core/Cargo.toml
@@ -50,3 +50,4 @@ aerospike = { path = "../" }
 aerospike-macro = {path = "../aerospike-macro"}
 lazy_static = "1.4"
 tokio = { version = "1.48.0", features = ["full"] }
+webpki-roots = "1"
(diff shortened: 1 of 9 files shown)


khaf (commented):
Why have you added the 'no_run' attribute? Don't the examples run correctly?

gagan405:
> Why have you added the 'no_run' attribute? Don't the examples run correctly?


Sync client also needs an async runtime. I get the below error while running those:

```
thread 'main' (1618547) panicked at /Users/gmishra/dev/aerospike-client-rust/aerospike-core/src/cluster/mod.rs:307:22:
there is no reactor running, must be called from the context of a Tokio 1.x runtime
stack backtrace:
```

In the latest commit I have updated the dependency and removed `no_run`s. Now it runs all doc tests throughout the project.

khaf (approved):
Looks good.

PR #195, opened 2026-04-02
Client 4405 execute query failing during nodechurn

dmishra-aero:
**Node Removal Process:** 
Before fix: 
   -    Remove alias 
   -    close node
   -   update  active node list

After fix:-
---------
- Make node inactive during discovery phase for node eviction
- Closing connection pool, when drop trait is invoked and that happen when all strong refs are gone

- set write policy - 0 by default
- honouring no-retries condition


khaf (approved):
Looks good.

PR #200, opened 2026-05-21
CLIENT-4751 change client timout from 30s to 1s, mirroring parity wit…

dmishra-aero:
In Rust client, Tend info call wait timeout by default `30s` causing longer wait time for timeout on a bad host
Java client default timeout is `1s` during tend info call


dmishra-aero:
Declining this PR in favour of changes that will be done by @khaf

PR #201, opened 2026-06-01
CLIENT-4881: Absorb all ServerError result codes onto the BatchRecord…

paul-brackin:
… instead of propagating them, and re-inject KeyNotFoundError from DeleteCommand.existed so delete-of-nonexistent reports the correct per-record code.


khaf (approved):
Looks good.

PR #202, opened 2026-06-04
CLIENT-4891 fix MessagePack string encoding for str8 range

gagan405:
For str8 range, the encoding is supposed to be `0xd9` followed by the length. This matches to how it is done in Java SDK or C/GO clients.

ref: https://github.com/aerospike/aerospike-client-java-sdk/blob/35ed5d3dacd802d4a1a43b68a864e32b4ac6bcc4/client/src/main/java/com/aerospike/client/sdk/util/Packer.java#L515-L517


khaf (approved):
Looks good.

PR #203, opened 2026-06-08
CLIENT-4878 serverside ael parsing

gagan405:
Adds functions to pack AEL request on server side. The AEL on server side uses the type 43 with a struct of ARRAY [128,  "ael string"].

This is to be consumed by python SDK.

Tested against a server that supports AEL parsing.


aerospike-core/src/cluster/version_parser.rs
```diff
@@ -209,4 +216,11 @@ mod tests {
             assert!(VersionParser::new(iv).parse().is_err());
         }
     }
+
+    #[test]
```
khaf, line 228: Unnecessary test.

khaf:
I took a brief look at the changes and nothing specific stood out to me. But we need the integration testing against the server before the eventual merge.

khaf (approved):
Approving for now, pending later changes.

PR #206, opened 2026-06-16
CLIENT-4973: Add positional Record.results: Vec<Value> for op-ordered access

paul-brackin:


aerospike-core/src/record.rs
```diff
@@ -35,9 +35,16 @@ pub struct Record {
...
-    /// Map of named record bins.
+    /// Map of named record bins. Convenient by-name access; collapses multiple ops
+    /// on the same bin via `Value::MultiResult` and drops nil-result ops.
     pub bins: HashMap<String, Value>,
 
+    /// Positional results, one slot per op in wire-arrival order. Preserves cross-bin
+    /// ordering and includes `Value::Nil` slots for ops that produced no value
+    /// (e.g. write-only modifies), so the index of a slot matches the index of its
+    /// originating op in the request's operation list.
+    pub results: Vec<Value>,
```
khaf: Not all records will have operation results. Having this allocated every time incurs significant performance penalty. Should be wrapped in `Option`.

aerospike-core/src/commands/read_command.rs
```diff
@@ -80,6 +80,7 @@ impl<'a> ReadCommand<'a> {
         expiration: u32,
     ) -> Result<(Record, Option<u64>)> {
         let mut bins: HashMap<String, Value> = HashMap::with_capacity(op_count);
+        let mut results: Vec<Value> = Vec::with_capacity(op_count);
```
khaf: This should be put in a `OnceCell` and only allocated for operations. Otherwise, be `None`. Not all commands are operations and that would incur significant performance penalties.

aerospike-core/src/commands/stream_command.rs
```diff
@@ -112,6 +112,7 @@ impl StreamCommand {
         }
 
         let mut bins: HashMap<String, Value> = HashMap::with_capacity(op_count);
+        let mut results: Vec<Value> = Vec::with_capacity(op_count);
```
khaf: Same here: `OnceCell`

khaf (changes requested):
The current implementation has performance implications that need to be addressed.

khaf:
I'm still not convinced regarding the performance characteristics of this feature. We'll need to talk about it before we merge it.

khaf (approved):

PR #209, opened 2026-06-23
[CLIENT-2388] Add a circuit breaker (max_error_rate, error_rate_window)

dmishra-aero:


aerospike-core/src/cluster/mod.rs
```diff
@@ -303,6 +309,21 @@ impl Cluster {
...
         let remove_list = self.find_nodes_to_remove(refresh_count).await;
         self.remove_nodes_and_aliases(remove_list);
+        // Bump the tend counter and, if the configured `error_rate_window`
+        // boundary lands here, roll every node's per-window error counter
+        // forward. Mirrors Java's
+        // `if (tendCount % errorRateWindow == 0) … resetErrorRate()`.
+        let tend_count = self
+            .tend_count
+            .fetch_add(1, Ordering::Relaxed)
+            .wrapping_add(1);
```
khaf, line 319: Isn't the `wrapping_add` redundant here?

dmishra-aero: `fetch_add(1)` increments the atomic, but it returns the previous value. `wrapping_add(1)` is just there to get the new value and keeps the same wrapping behavior if the counter reaches `usize::MAX` so it's not redundant here

PR #213, opened 2026-06-29
CLIENT-4989: classify socket I/O errors as Connection err, add pool liveness probe

dmishra-aero:


dmishra-aero:
@khaf does v3 support `rt-async-std` ? If the mandate is use `tokio` runtime do we need test running against this `async-std` runtime in CI-CD pipeline..
Build are failing due to that ..

khaf:
Is was due to merging a branch from Paul. I have fixed the issues, so you should be able to pull the changes in.

khaf (approved):
Looks good.

PR #214, opened 2026-07-01
CLIENT-4751: included different timeout used across different ops

dmishra-aero:
### Out of Scope: `resetDeadline` parity for explicit `connect_timeout`

This PR intentionally does not implement Java's `resetDeadline` behavior for newly established pooled connections.

In Java, when `Policy.connectTimeout > 0`, the command `total_timeout` deadline is reset after connection establishment (including TCP connect, TLS handshake, and authentication), ensuring connection setup time does not consume the command SLA.

This is deferred because `BasePolicy.connect_timeout` defaults to `0`, so Java does not invoke `resetDeadline` in the default path, and steady-state traffic typically uses existing pooled connections. The focus of this PR is timeout separation (`ClientPolicy.timeout`/`login_timeout` vs. command timeouts), not full command deadline accounting parity.

The difference is only observable when users explicitly configure `connect_timeout > 0` together with a tight `total_timeout` (for example, TLS handshakes on pool misses).

This can be added later via a small follow-up change in `SingleCommand` and batch command paths as part of separate ticket

### DIFF
--- a/aerospike-core/src/cluster/mod.rs
+++ b/aerospike-core/src/cluster/mod.rs
@@ -1,4 +1,4 @@
-// Copyright 2015-2024 Aerospike, Inc.
+// Copyright 2015-2026 Aerospike, Inc.
 //
 // Portions may be licensed to Aerospike, Inc. under one or more contributor
 // license agreements.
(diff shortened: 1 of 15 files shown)


khaf:
Decided against this PR in a meeting with @dmishra-aero . It has been fixed in a different branch.

PR #215, opened 2026-07-02
CLIENT-5032: Add append/prepend string operations (sub-ops 67/68)

paul-brackin:


khaf:
Done  in a different changeset.

PR #218, opened 2026-07-06
CLIENT-4966: restore connection to  Ready state after authenticate on Ok path

dmishra-aero:
## Root Cause

`authenticate()` and `authenticate_with_session()` set the connection state to Writing on entry, but failed to restore it to `Ready` on the successful (`Ok`) path. As a result, newly created Connection instances remained in the Writing state.

When `fill_min_conns()` invoked `make_conn().await`?, the returned connections were dropped as temporaries. Because their state was still `Writing`, `PooledConnection::Drop` treated them as non-recoverable and closed them instead of returning them to the pool. Consequently, the pool's reserved connection count never increased, causing every subsequent tend cycle to recreate the full `min_conns_per_node` set indefinitely.

Three regression tests have been added to verify the pool's `Drop` contract and prevent this behavior from regressing.

Notably, this bug occurs regardless of whether authentication actually takes place. The connection state is set to `Writing` before the authentication mode is checked, so the incorrect state persists even when no authentication bytes are sent over the wire.

## Test Environment

* **Aerospike Enterprise Edition:** 8.1.2.3, 3-node cluster via aerolab
* **Client:** This branch, benchmark tool
* **`min_conns_per_node`:** 32
* **`conn_pools_per_node`:** 4
* **Workers (`-t`):** 8 (< `min_conns_per_node`)
* **Workload:** Mixed reads and writes
* **Duration:** 55 seconds
* **Observation Method:** `docker exec asinfo -v statistics` polled every second

## Results: Before vs. After the Fix

The benchmark was executed twice using identical code, commands, and cluster configuration: once on the unfixed branch and once after applying the fix.

* **Connection opens per second:** 36 → 1
* **Total connections opened (~44 s):** 1,061 → 35
* **Server-side live connection count:** Oscillating between 10 and 34 → Stable at 34
* **Operations per second:** ~9,700 → ~9,970
* **Errors:** 0 → 0
* **Timeouts:** 0 → 0

### DIFF
--- a/aerospike-core/src/net/connection.rs
+++ b/aerospike-core/src/net/connection.rs
@@ -685,7 +685,12 @@ impl Connection {
     ) -> Result<Option<crate::commands::admin_command::SessionInfo>> {
         self.state = ConnectionState::Writing;
         match AdminCommand::authenticate(self, auth_mode, hashed_pass).await {
-            Ok(session) => Ok(session),
+            Ok(session) => {
+                // Restore Ready so PooledConnection::Drop puts the conn back
+                // in the pool instead of taking the non-recoverable close arm.
+                self.set_state(ConnectionState::Ready);
+                Ok(session)
+            }
             Err(err) => {
                 self.close();
                 Err(err)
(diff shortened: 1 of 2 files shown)


khaf (approved):
Looks good.

khaf:
@dmishra-aero Does this need to be back-ported to the v2?

PR #219, opened 2026-07-09
CLIENT-4966: restore connection to Ready state after authenticate on OK path

dmishra-aero:
Fix: authenticate() was setting connection state to Writing but failing to restore it to Ready on successful completion. As a result, connections were dropped by PooledConnection::Drop instead of being returned to the pool, causing fill_min_conns() to endlessly recreate connections (~36 opens/sec).

With this fix, connections are properly recycled (Ready state), eliminating the churn fixpoint and reducing connection creation to ~1 open/sec.

Changes:
- authenticate() now calls set_state(Ready) on Ok path
- Added 3 regression tests validating the pool's Drop contract

Metrics:
- Connection opens/second: 36 → 1 (36x reduction)
- Total new connections (44s): ~1,061 → ~35 (30x reduction)
- Server-side connection stability: Oscillating → Stable

### DIFF
--- a/aerospike-core/src/net/connection.rs
+++ b/aerospike-core/src/net/connection.rs
@@ -60,6 +60,7 @@ pub enum ConnectionState {
 
 /// Underlying socket type for a connection (TCP or TLS).
 #[derive(Debug)]
+#[cfg_attr(test, allow(dead_code))]
 #[allow(clippy::large_enum_variant)]
 pub enum Netsocket {
     /// Plain TCP stream.
@@ -95,6 +96,7 @@ pub struct Connection {
 }
 
 impl Connection {
+    #[cfg_attr(test, allow(dead_code))]
     #[cfg(feature = "tls")]
     async fn get_netsocket(
         stream: TcpStream,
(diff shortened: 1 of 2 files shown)


tests/src/connection_churn_bench.rs
paul-brackin, line 1: This is a heavy addition to mod.rs, which will run it on every `cargo check`. It has fixed sleeps, live-cluster dependency, lots of println!, and few assertions.  Consider removing from mod.rs or #[ignore]/feature-gate this one.  Prolly belongs more in the QA test suite, so we could advise Diane to build something similar for the PAC if she hasn't already.

tests/src/connection_churn_metrics.rs
paul-brackin, line 1: Test can return successfully when no cluster is reachable, so the test can go green without validating the fix.   Its also heavy for a run on every `cargo check`.  Consider removing from mod.rs or #[ignore]/feature-gate this one as well.  Prolly belongs more in the QA test suite, so we could advise Diane to build something similar for the PAC if she hasn't already.

paul-brackin (changes requested):
Looks good overall... Just some thought on those 2 new tests...

paul-brackin (approved):
Looks good

PR #220, opened 2026-07-09
CLIENT-4989: classify socket I/O errors as Connection err

dmishra-aero:


aerospike-core/src/net/connection_pool.rs
```diff
@@ -121,6 +121,14 @@ impl Queue {
                     // let the connection drop and close
```
paul-brackin, line 121: Queued conns hold a reserved slot, so evicting one must call reduce_capacity(). The idle branch does not.  Consider adding `self.reduce_capacity()` here

aerospike-core/src/net/connection.rs
```diff
@@ -467,6 +468,35 @@ impl Connection {
...
+                #[cfg(feature = "rt-async-std")]
+                {
+                    // async-std's `TcpStream` exposes no non-blocking probe;
+                    // conservatively treat as alive.
+                    let _ = s;
+                    true
+                }
+            }
+            #[cfg(feature = "tls")]
+            Netsocket::Tls(_) => true,
```
paul-brackin, line 494: Seems like we are neglecting the Enterprise case here?  Would something like this work for TLS as well?:

`recv(fd, buf, MSG_PEEK | MSG_DONTWAIT)` (on the underlying TCP file descriptor)

aerospike-core/src/net/connection.rs
```diff
@@ -238,7 +238,8 @@ impl Connection {
 
         match res {
             Ok(Ok(())) => (),
-            Ok(Err(e)) => return Err(e.into()),
+            // classify socket I/O errors as Connection err and hence command retries.
+            Ok(Err(e)) => return Err(Error::Connection(format!("flush: {e}"))),
```
paul-brackin, line 242: This reclassifies socket I/O failures as Error::Connection, which makes them retriable through commands::is_network_error() / should_retry(). This applies to all reclassified arms in this PR: flush, read, write, read_all, and buffered_read.

Read-side failures look safe, but write-side failures (flush / write) can be in-doubt if bytes already reached the server. Can you confirm the command retry layer still prevents unsafe retries for non-idempotent writes and preserves expected in-doubt behavior?

paul-brackin (changes requested):

aerospike-core/src/net/connection.rs
dmishra-aero: The current `NetSocket::Tls(_) => true` behavior is intentional and matches the Java client. I verified that `NioConnection.isValid()` performs a 1-byte non-blocking peek for plain TCP, while `NettyConnection.isValid() (TLS)` simply returns `channel.isActive()` without a byte-level probe.

Your suggestion of using `recv(..., MSG_PEEK | MSG_DONTWAIT)` on the underlying socket would provide a stronger liveness check for TLS than Java currently has. However, we can't use `TcpStream::try_read()` since it consumes bytes instead of peeking, which could interfere with `rustls`. A proper implementation would require `MSG_PEEK` via `socket2` crate or a raw `libc` call.

Since the scope of this PR is limited to fixing the reported bug and adding the plain TCP liveness check, I would prefer to address the stronger TLS liveness check in as a seperate jira provided if we agree it's worth adding.

aerospike-core/src/net/connection.rs
dmishra-aero: Thanks for raising this. I dug into the `v2` retry loop and compared it with `v3` and `Java`.
The key finding is that this PR does not change the write retry behavior.

In `v2`, errors returned from `write_buffer()` are already retried unconditionally:

```
if let Err(err) = cmd.write_buffer(&mut conn).await {
    ...
    last_err = Some(err);
    continue;
}
```
This path never consults `should_retry()`, so changing the error classification from `Error::Io` to `Error::Connection` has no effect on write retries:

**Before** this PR: `Error::Io → retry`
**After** this PR: `Error::Connection → retry`

The classification change only affects the `read` path, where `should_retry()` determines whether a retry should occur.
While investigating, I also noticed that `v2's` single-command path does not implement the `in_doubt` mechanism that already exists in `Java` and `v3`

So bottomline is changes only affect the `read` path

aerospike-core/src/net/connection.rs
paul-brackin: Yeah this maybe a case where we can improve beyond the reference client.  Wouldnt be the first.  But I'm ok limiting the scope here.

dmishra-aero: Better to deal with this seperately

aerospike-core/src/net/connection.rs
paul-brackin: Clarification note:  "changes only affect the read path"; this is almost true... a failed response read of a write command is the in-doubt case as that path IS newly retriable.  Its just defaulted to max_retries = 0, so only matter when the user opts in there.  Otherwise sounds good.

paul-brackin (approved):

PR #221, opened 2026-07-10
CLIENT-5011 server side idx selection

gagan405:
Add wire protocol handling for index selection on server side.
The base branch is setup to `CLIENT-4878-serverside-ael-parsing` as index selection will only work with server side AEL


aerospike-core/src/commands/query_explain_command.rs
```diff
@@ -0,0 +1,171 @@
...
+            result_code,
+            &self.namespace,
+            self.set_name.as_deref(),
+            self.explain_where_bytes.clone(),
+            &fields,
+        )?);
+
+        // Single-message response fully consumed; do not call empty_socket (it
+        // re-reads the buffer as a proto header and mis-parses field TLV bytes).
+        conn.buffer.data_buffer.clear();
```
khaf: This can potentially be expensive. Why is it necessary?

aerospike-core/src/query/plan/query_plan.rs
```diff
@@ -0,0 +1,301 @@
...
+    set_name: Option<String>,
+    explain_where_bytes: Vec<u8>,
+    index_name: Option<String>,
+    index_range_bytes: Option<Vec<u8>>,
+    index_type: CollectionIndexType,
+}
+
+impl QueryPlan {
+    /// Builds a plan from an explain `result_code` and parsed response fields.
+    pub fn from_explain_response(
```
khaf: Should be `pub(crate)`. Not needed for the end user.

aerospike-core/src/query/plan/query_plan.rs
```diff
@@ -0,0 +1,301 @@
...
+        self.set_name.as_deref()
+    }
+
+    /// AEL source text from the explain field `44` payload.
+    pub fn ael(&self) -> Result<String> {
+        QueryWhereWire::ael(&self.explain_where_bytes)
+    }
+
+    /// Field `44` body sent on explain (`EXPLAIN` flag set).
+    pub fn explain_where_bytes(&self) -> &[u8] {
```
khaf: Should be `pub(crate)`

aerospike-core/src/query/plan/query_plan.rs
```diff
@@ -0,0 +1,301 @@
...
+        &self.explain_where_bytes
+    }
+
+    /// Field `44` body for execute (`EXPLAIN` flag cleared).
+    ///
+    /// Clears the explain flag in place and returns the payload buffer.
+    pub fn into_execute_where_bytes(mut self) -> Vec<u8> {
```
khaf: Should be pub(crate)

aerospike-core/src/commands/buffer.rs
```diff
@@ -1474,7 +1475,7 @@ impl Buffer {
                 field_count += 1;
             }
 
-            filter_size = 1 + filter.estimate_size()?;
+            filter_size = filter.index_range_field_body_size()?;
```
khaf, line 1478: Why did the name change?

gagan405: The old `1 + filter.estimate_size()` was only correct for client-built bin filters. Server side selection also uses `INDEX_RANGE` bytes from explain. So I created this new fn that covers both branches.

aerospike-core/src/commands/query_explain_command.rs
gagan405: It is redundant and was an extra unnecessary guard. Removing it.

khaf:
Why was this merged without my go ahead?

PR #222, opened 2026-07-14
Fix batch retry timeout error context

dmishra-aero:
**Issue:** 

**warning**: value assigned to `last_err` is never read
```
  --> aerospike-core/src/commands/batch_operate_command.rs:75:43
   |
75 |         let mut last_err: Option<Error> = None;
   |                                           ^^^^
   |
   = help: maybe it is overwritten before being read?
   = note: `#[warn(unused_assignments)]` (part of `#[warn(unused)]`) on by default

```

## Fix's Summary 
This change fixes the warning while preserving the existing functional behavior. When batch operate retries are exhausted, the client continues to return a timeout as the primary error, but now also preserves the last retry failure by wrapping it in the existing Error::Chain. As a result, underlying failures (such as circuit-breaker backoff or max error rate) remain visible in application logs instead of being obscured by a generic timeout.

## Semantic Impact
Before this change, batch retry exhaustion returned only `Error::Timeout`, losing the final underlying error. After this change, the displayed error includes both the timeout and the last retry failure, e.g. `Client Timeout: Timeout after ...` followed by `Max error rate exceeded ...`.

The public error type is unchanged. Since the returned value is now `Error::Chain(Timeout, last_error)` on these batch paths, callers that directly match only `Error::Timeout(_)` may need to account for `chained` errors.


aerospike-core/src/commands/batch_operate_command.rs
```diff
@@ -380,3 +388,40 @@ impl BatchOperateCommand {
         Ok(())
     }
 }
+
```
khaf: These tests are just slop. Please remove them.

khaf (changes requested):

PR #223, opened 2026-07-14
Client 4751 connect timeout

khaf:
@dianetong This branch includes the fix for https://aerospike.atlassian.net/browse/CLIENT-4751


khaf:
Already done in a different change.

PR #224, opened 2026-07-16
CLIENT-5033: Fix client hang on compressed responses > 30 bytes

dmishra-aero:
Use `read_buffer_at(already_read, remaining)` instead of `read_buffer(remaining) `
to prevent overwriting proto header on second read, fixing possible client hangs on 
compressed responses > 30 bytes.

### DIFF
--- a/aerospike-core/src/net/connection.rs
+++ b/aerospike-core/src/net/connection.rs
@@ -500,18 +500,14 @@ impl Connection {
         // Need to read more data from the network
         let remaining = total_message_size - already_read;
 
-        // Save what we already have after the 16-byte compressed header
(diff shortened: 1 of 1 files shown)


khaf:
@dmishra-aero What's the status of this PR. Is it confirmed and ready to go? If so, can you please add tests to it?

dmishra-aero:
> @dmishra-aero What's the status of this PR. Is it confirmed and ready to go? If so, can you please add tests to it?

Hey, I had to pause this work for now, but will get back to it soon once I wrap up my current work.

dmishra-aero:
I cross checked it, and it looks like my initial assessment no longer holds true. I also double-checked that the get/read compressed-response path is correctly verified at the connection layer through tests.
hence i am closing it

PR #230, opened 2026-07-31
Change vector to struct for future parameters

DKZed:


aerospike-core/src/vector.rs
```diff
@@ -396,35 +493,79 @@ mod tests {
     use super::*;
     use crate::commands::buffer::Buffer;
 
+    /// Write `vector` into a fresh buffer sized to (at least) `capacity` and
+    /// reset the offset to the start, ready to read back.
+    fn encode(vector: &Vector, capacity: usize) -> Buffer {
+        let mut buf = Buffer::new(usize::MAX);
```
vkagamlyk, line 499: is it possible to create buffer with size `capacity`?

DKZed: 1. This is test code so it's mostly throwaway (usize::MAX is default /no-op)
2. This is the reclaim-size memory optimization when and if buffer is reduced - not buffer creation size

vkagamlyk (approved):

PR #233, opened 2026-08-15
CLIENT-5202: evict dead pooled sockets under default policy (idle_timeout=0)

dmishra-aero:
The pool is now `LIFO`: `put_back()` still pushes to the **back**, and `get()`
now pops from the **back** too. The last connection returned is the first
one lent out, so a few hot connections serve the traffic while the
unused ones age at the front and retire on schedule.

**Note**:  We could have implemented **LIFO** using either end of the `VecDeque`, as long as `get() `and `put_back()` use the same end. I chose to keep the hottest connection at the back and the oldest at the front to minimize code changes. This is equivalent in both behavior and performance, since `VecDeque` supports `O(1)` operations at both ends. The important part is that `get()` always picks the most recently returned `(hottest)` connection first.

## Benchmark against `V3` as baseline and observation are captured below:
- **Hot path unchanged** — sustained 45 s ×2: 9,380/9,348 tps (v3) vs 9,418/9,392 tps; p99 ~1.1 ms both. Within noise.
- **No cold-tail penalty** — burst after 90 s idle: 0 reconnects both sides; p99 1.27 ms (v3) vs 1.12 ms.
 - **Cyclic rebuild cost inherited, not introduced** — peak/valley with idle=2s, min=4: +12 conns/cycle on both sides; p99 equal.
- **Default policy stays stable** — peak/valley with idle=0: pool built once (+15), then +0 forever, both sides.
No churn at rest — 180 s idle, min=8: both settle 16→8, 0 opened.

### DIFF
--- a/aerospike-core/src/cluster/node.rs
+++ b/aerospike-core/src/cluster/node.rs
@@ -14,6 +14,7 @@
 // the License.
 
 use std::collections::HashMap;
+use std::time::Duration;
 use indexmap::IndexMap;
 use std::fmt;
 use std::hash::{Hash, Hasher};
@@ -33,7 +34,7 @@ use crate::cluster::CLIENT_VERSION;
 use crate::commands::Message;
 use crate::errors::{Error, Result};
 use crate::metrics::NodeMetrics;
-use crate::net::{Connection, ConnectionPool, Host, PooledConnection};
+use crate::net::{Connection, ConnectionPool, Host, PooledConnection, Queue, TailVerdict};
 use crate::policy::{AdminPolicy, ClientPolicy};
 use crate::Version;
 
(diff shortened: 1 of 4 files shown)


khaf:
Had a discussion with @dmishra-aero regarding the algorithm. The pool is supposed to be a Heap (LIFO) to lend itself to implementing the idle connection management, while at the moment it is a Queue (FIFO). We decided that it needs to change and all the algorithms to be adapted to that change.

aerospike-core/src/cluster/node.rs
```diff
@@ -916,91 +939,86 @@ impl Node {
...
-                queue.reduce_capacity();
-                self.metrics.incr_connections_idle_dropped();
-                self.metrics.incr_connections_closed();
-                total_processed += 1;
-            }
-            let keepers: Vec<Connection> = idle_iter.collect();
-
-            if keepers.is_empty() {
-                continue;
+        let mut active: std::collections::VecDeque<&Queue> =
```
khaf: This allocation here is unnecessary. We can use `.cycle().iter()` to cycle over the sub-pools.

khaf (changes requested):
We need to avoid unnecessary allocation as much as possible, unless otherwise the algorithm becomes too complex for what it does.

khaf (approved):
Looks good.

PR #234, opened 2026-08-18
[CLIENT-5202] Connection Liveness Check

khaf:
PR #227 removed Connection::is_alive() and its guard in Queue::get() on the assumption that Node::reap_and_refresh_idle_connections() would evict stale sockets during tend. It does not under default policy: idle_timeout defaults to 0, so idle_deadline is never set, is_idle() always answers false, and try_extract_idle() hands the reaper an empty list. With no checkout-side check either, sockets a restarted server has already closed stay in the pool until a command draws one and fails on its first read with `early eof`, burning a retry.

The probe is back, in the form v2 arrived at: a non-blocking one-byte MSG_PEEK through socket2's SockRef, which borrows the fd without owning it and needs no unsafe. Unlike the original try_read version it consumes nothing, so a byte it sees is still there for the command that follows. It is deliberately independent of idle_timeout, because a socket can die long before its idle deadline — a server restart kills sockets that were in use a millisecond earlier.

Coverage now extends past plain TCP on tokio. The same peek serves the TcpStream under a rustls session and async-std's socket, both of which previously had no liveness check at all. TLS differs in one rule: pending bytes count as alive there, because post-handshake control records (TLS 1.3 NewSessionTicket, KeyUpdate) legitimately arrive while a connection sits idle in the pool, and evicting on them would reconnect, receive a fresh ticket and evict again — churn caused by the probe itself. Only a closed or broken socket is fatal.

Evicting at checkout counts as a closed connection in the node metrics. aerospike-rt enables async-std's "io_safety" feature for its AsFd impls.

### DIFF
--- a/aerospike-core/Cargo.toml
+++ b/aerospike-core/Cargo.toml
@@ -41,6 +41,10 @@ regex = "1"
 tokio-rustls = { version = "0.26.4", optional = true }
 rustls = { version = "0.23.42", optional = true }
 async-channel = "2.5.0"
+# Borrowed, non-owning socket handle for the connection pool liveness probe:
+# one implementation covering tokio, async-std and the TLS inner stream, with no
+# unsafe and no fd ownership. Already in the tree via tokio.
+socket2 = "0.6"
 hazarc = "0.2.0"
 indexmap = "2"
 # `async` so stream reads/writes are await points on the client executor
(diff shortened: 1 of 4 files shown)


dmishra-aero (approved):
Look's good to me..

PR #235, opened 2026-08-18
V2 improvements

khaf:


aerospike-core/Cargo.toml
```diff
@@ -33,12 +33,19 @@ regex = "1"
...
 
 [features]
 serialization = ["serde"]
-rt-tokio = ["aerospike-rt/rt-tokio"]
-rt-async-std = ["aerospike-rt/rt-async-std"]
+# The aerospike-macro features pick the runtime for `#[aerospike_macro::test]`.
+# They used to come in via the `aerospike` facade dev-dependency, which also
+# forced rt-tokio and made rt-async-std test builds impossible.
+rt-tokio = ["aerospike-rt/rt-tokio", "aerospike-macro/rt-tokio"]
+rt-async-std = ["aerospike-rt/rt-async-std", "aerospike-macro/rt-async-std"]
```
dmishra-aero, line 48: **cargo check --features sync,rt-tokio** failing with below

`Error[E0425]: cannot find function block_on in this scope
   --> aerospike-sync/src/client.rs:157:22
    |
157 |         let client = block_on(aerospike_core::Client::new(policy, hosts))?;
    |                      ^^^^^^^^ not found in this scope 
`

Forward rt-tokio / rt-async-std to aerospike-sync (aerospike-sync?/rt-tokio). Without it, features = ["sync", "rt-tokio"] does not compile; the sync client fixes in this PR are unreachable via the facade.

aerospike-core/src/value.rs
```diff
@@ -239,8 +251,24 @@ impl Value {
...
+    /// reports the same case as `PARAMETER_ERROR` from `Value.getType()`.
+    ///
+    /// This used to be `unreachable!()`, so either value killed the process.
+    /// The panic is the return path here: there is no `ParticleType` that
+    /// safely means "has no particle type", and a placeholder would trade a
+    /// loud abort for a silent buffer-size mismatch, since `estimate_size`
+    /// reports 0 for these while `write_to` packs real bytes. Failing in
+    /// `particle_type`, which every particle path calls before `write_to`, is
+    /// what keeps that unreachable.
+    pub fn particle_type(&self) -> Result<ParticleType> {
```
dmishra-aero: As this is a public API, I assume it was intended primarily for internal use. If that's not the case, changing the outcome type could be a breaking change. Are we planning to bump the major version for this change, or can we assume that end users aren't expected to use this API since it was originally intended for internal purposes?

aerospike-core/src/value.rs
```diff
@@ -90,6 +90,18 @@ impl From<&f64> for FloatValue {
     }
 }
 
+impl FloatValue {
+    /// Widen to `f64` for the wire, where every float is an 8-byte double
+    /// particle. `f32 -> f64` is lossless (matches the Java client, which
+    /// stores both Float and Double bins as doubles).
+    pub(crate) const fn as_f64(&self) -> f64 {
```
dmishra-aero, line 88: I might be missing context but i noticed after the `f32` bin change, is there a reason `From<FloatValue>` for `f64` still rejects F32 while other code paths accept it via `as_f64()`? I’m trying to understand whether that’s an intentional guardrail or something we should align before merge.

aerospike-core/src/net/connection.rs
```diff
@@ -407,11 +489,11 @@ impl Connection {
         let timeout = self.deadline();
         let res = match self.conn {
             Netsocket::Tcp(ref mut conn) => {
-                aerospike_rt::timeout(timeout, conn.write_all(buf)).await
+                io_with_timeout!(self, timeout, conn.write_all(buf))
             }
             #[cfg(feature = "tls")]
             Netsocket::Tls(ref mut conn) => {
-                aerospike_rt::timeout(timeout, conn.write_all(buf)).await
+                io_with_timeout!(self, timeout, conn.write_all(buf))
```
paul-brackin: CLIENT-5268 flushed the rustls session in `Connection::flush()` (line 327) but not here. `write_all` still does `conn.write_all(buf)` with no `flush()`, so ciphertext can sit in the TLS session buffer until a later write.

That is the tend/info path: `info_command.rs:66` calls `conn.write_all(&self.buf)` and then immediately `read_all`s the reply. Admin and KV commands go through `flush()` and are fine; TLS tend is not.

`write_all` should flush the TLS session the same way `flush()` does; otherwise CLIENT-5268 only covers one of the two write entry points.

aerospike-core/src/commands/batch_operate_command.rs
```diff
@@ -88,22 +108,44 @@ impl BatchOperateCommand {
...
+                let mut nodes: Vec<Arc<Node>> = Vec::with_capacity(self.batch_ops.len());
+                for ((op, _), previous) in self.batch_ops.iter().zip(&last_tried) {
+                    let partition = Partition::new_by_key(op.key());
+                    nodes.push(cluster.get_node(
                         &partition,
                         self.policy.replica,
-                        Some(self.node.clone()),
-                    )?;
+                        Some(previous.clone()),
+                    )?);
```
paul-brackin: This retry still uses ?, so an un-routable key fails far more than this command. execute_batch_operate_jobs collects Vec<Result<_>> into Result<Vec<_>> and batch_execute does .await? on it, so the first Err short-circuits and discards the completed results from every other node's command too & the whole batch is lost.

That's the same failure mode the first-attempt split just removed ("Propagating here discarded every other key's result before anything was even sent"), one level down and after the work is done. It's also most likely exactly here: you're on a Sequence/PreferRack retry because a node went away, which is the same condition that makes get_node fail.

Stamp the failed key PartitionUnavailable and continue, matching the first-attempt split; only fail outright when nothing is routable.

aerospike-core/src/query/recordset.rs
paul-brackin: poll_next still busy-wakes: try_recv + wake_by_ref() + Pending spins hot whenever the channel is empty and the query is active. The new Closed arm only stops the spin after close(). The blocking iterator now parks via recv_blocking(); the async stream should park too, via recv()/poll_recv.

paul-brackin (changes requested):
I ran some local PAC testing against it & it was mostly good. But...

Agree with Deeraj's findingss + see my inline review comments mostly from my AI agent.  Also - just a note that the batch change is downstream-visible. An un-routable key used to raise InvalidNode for the whole batch and now yields a per-key PartitionUnavailable with the batch succeeding.

PR #239, opened 2026-08-21
CLIENT-5228 error verbosity on query plan

gagan405:
* enable propagating error verbosity flag in query plan phase
* added tests


khaf (approved):
Looks good.

PR #250, opened 2026-09-01
[CLIENT-5292]: add test coverage check for aerospike-core

diparthaspike:
Fixes: https://aerospike.atlassian.net/browse/CLIENT-5292
## What this adds

A `cargo-llvm-cov`-based coverage check for `aerospike-core`, running as its own job (`aerospike-core-coverage`) on every PR. Coverage tooling did not exist in this repo before this PR.

## Why a separate check, not folded into the existing `test` job

Two different concerns, two different signals:
- `test (rt-tokio / rt-async-std)` — unchanged, still fails on any test failure.
- `aerospike-core-coverage` — a new, independent check that only asks "is the coverage floor met?"

Keeping them separate means a red X tells you *which* thing failed at a glance, instead of one job that can go red for two unrelated reasons. It also runs in parallel with the two `test` jobs rather than adding to either one's runtime.

## Cost

Effectively zero. This is a public repo, so GitHub Actions minutes are free and unlimited on standard runners regardless of account type — this PR adds one extra job, not a new spending category. The only real cost is wall-clock time (roughly 2-5 min per PR run), and it runs concurrently with the existing jobs, not after them.

## Why `rt-tokio` only, not both runtimes

Coverage is collected once, using the `rt-tokio` feature set (the default runtime per `Cargo.toml`), not once per runtime. The code paths being measured aren't runtime-specific — running the full suite twice under instrumentation to cover both `rt-tokio` and `rt-async-std` would double the time for no meaningful extra signal.

## Why coverage still gets computed even if tests fail

Both collection steps use `--ignore-run-fail`. Coverage instrumentation and test pass/fail are two separate systems: a test can fail an assertion and still have exercised (and gotten credit for) every line it touched before failing. Without `--ignore-run-fail`, `cargo-llvm-cov` refuses to emit a report at all if any test fails — which would mean one flaky or environment-specific failure silently blocks the coverage signal entirely. Test correctness is still fully enforced elsewhere, by the existing `test` job, which is untouched.

## Denominator: `aerospike-core` only

The coverage percentage is reported against `aerospike-core` specifically, not the root `aerospike` crate. `aerospike` is an 11-line re-export facade — scoping the report to it produces a near-empty, meaningless number. `aerospike-core`'s own inline unit tests and the integration/proptest suite (which runs via the `aerospike` package) are collected separately and merged into one report scoped to `aerospike-core`, so both test tiers count toward the same number.

## Threshold: 60% now, not 80%

Check this: https://github.com/aerospike/aerospike-client-rust/pull/250#issuecomment-5509012590

Note: the exact percentage varies slightly between runs — proptest generates randomized inputs per run (no fixed seed), so which branches get exercised isn't bit-for-bit identical every time. This is expected and inherent to property-based testing, not a flaky check.

## Doc tests: not counted

`aerospike-core` and `aerospike-sync` doc tests already run and must pass (existing steps, untouched) — they're just not included in this coverage percentage. Whether they should count is an open question, deliberately deferred, not decided here.

## Scope of this PR

Coverage tooling only. No test fixes, no coverage improvements, no changes to the existing `test` job.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -69,3 +69,44 @@ jobs:
           # aerospike-core doc examples use `use aerospike::…`; enable rt-tokio on the path dev-dep.
           sed -i 's/features = \["async"\]/features = ["async", "rt-tokio"]/' aerospike-core/Cargo.toml
           cargo test --doc -p aerospike-core --features rt-tokio,serialization
+
+      - name: Install cargo-llvm-cov
+        if: matrix.runtime == 'rt-tokio'
+        uses: taiki-e/install-action@1ed6d7be6168f6c9046541087ff549b6bc581fdf # v2.87.2
+        with:
+          tool: cargo-llvm-cov
+
+      # Coverage denominator is aerospike-core only (see CLIENT-5292): the root
+      # `aerospike` crate is an 11-line re-export facade, and reporting against it
+      # instead of aerospike-core silently produces an empty/near-empty report.
+      # aerospike-core's own inline unit tests and the integration/proptest suite
+      # (run via the `aerospike` package) are collected separately, then merged
+      # into one report scoped to aerospike-core.
+      #
+      # Doc tests are intentionally excluded from this number for now (undecided,
+      # tracked on the ticket) even though they run and must pass above.
+      - name: Coverage - aerospike-core unit tests
+        if: matrix.runtime == 'rt-tokio'
+        env:
+          PROPTEST_MAX_SHRINK_ITERS: 0
+        run: |
+          cargo llvm-cov clean --workspace
+          cargo llvm-cov -p aerospike-core \
+            --no-default-features --features serialization,rt-tokio \
+            --ignore-run-fail --no-report
+
+      - name: Coverage - aerospike integration and property tests
+        if: matrix.runtime == 'rt-tokio'
+        env:
+          PROPTEST_MAX_SHRINK_ITERS: 0
+        run: |
+          cargo llvm-cov -p aerospike \
+            --no-default-features --features serialization,async,rt-tokio \
+            --ignore-run-fail --no-report -- --test-threads=1
+
+      # Soft floor while coverage tooling beds in: current aerospike-core baseline
+      # is ~79% (recorded on CLIENT-5292). 75% leaves headroom to catch a real
+      # regression without blocking on day one; raise to 80% once confirmed.
+      - name: Coverage report (aerospike-core)
+        if: matrix.runtime == 'rt-tokio'
+        run: cargo llvm-cov report -p aerospike-core --fail-under-lines 75
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -70,8 +70,33 @@ jobs:
           sed -i 's/features = \["async"\]/features = ["async", "rt-tokio"]/' aerospike-core/Cargo.toml
           cargo test --doc -p aerospike-core --features rt-tokio,serialization
 
+  aerospike-core-coverage:
+    name: aerospike-core-coverage
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Harden the runner (Audit all outbound calls)
+        uses: step-security/harden-runner@20cf305ff2072d973412fa9b1e3a4f227bda3c76 # v2.14.0
+        with:
+          egress-policy: audit
+
+      - name: Checkout
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+
+      - name: Install Rust 1.87 (MSRV)
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: 1.87
+
+      - name: Rust cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: aerospike-core-coverage
+
+      - name: Set up Aerospike Database
+        uses: reugn/github-action-aerospike@2065a9209cfd5ef88a3e07f3e7929e321d1e0067 # v1.1.0
+
       - name: Install cargo-llvm-cov
-        if: matrix.runtime == 'rt-tokio'
         uses: taiki-e/install-action@1ed6d7be6168f6c9046541087ff549b6bc581fdf # v2.87.2
         with:
           tool: cargo-llvm-cov
@@ -84,9 +109,8 @@ jobs:
       # into one report scoped to aerospike-core.
       #
       # Doc tests are intentionally excluded from this number for now (undecided,
-      # tracked on the ticket) even though they run and must pass above.
+      # tracked on the ticket) even though they run and must pass in the `test` job.
       - name: Coverage - aerospike-core unit tests
-        if: matrix.runtime == 'rt-tokio'
         env:
           PROPTEST_MAX_SHRINK_ITERS: 0
         run: |
@@ -96,7 +120,6 @@ jobs:
             --ignore-run-fail --no-report
 
       - name: Coverage - aerospike integration and property tests
-        if: matrix.runtime == 'rt-tokio'
         env:
           PROPTEST_MAX_SHRINK_ITERS: 0
         run: |
@@ -108,5 +131,4 @@ jobs:
       # is ~79% (recorded on CLIENT-5292). 75% leaves headroom to catch a real
       # regression without blocking on day one; raise to 80% once confirmed.
       - name: Coverage report (aerospike-core)
-        if: matrix.runtime == 'rt-tokio'
         run: cargo llvm-cov report -p aerospike-core --fail-under-lines 75
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -74,6 +74,10 @@ jobs:
     name: aerospike-core-coverage
     runs-on: ubuntu-22.04
 
+    env:
+      # 75% is the starting point to keep infra adoption moving -- nothing less.
+      COVERAGE_THRESHOLD: 75
+
     steps:
       - name: Harden the runner (Audit all outbound calls)
         uses: step-security/harden-runner@20cf305ff2072d973412fa9b1e3a4f227bda3c76 # v2.14.0
@@ -118,16 +122,14 @@ jobs:
             --no-default-features --features serialization,async,rt-tokio \
             --ignore-run-fail --no-report -- --test-threads=1
 
-      # 75% is the starting point to keep infra adoption moving -- nothing less.
       - name: Coverage report (aerospike-core)
         run: |
-          THRESHOLD=75
           cargo llvm-cov report -p aerospike-core
           PCT=$(cargo llvm-cov report -p aerospike-core --json --summary-only | jq -r '.data[0].totals.lines.percent')
           PCT_FMT=$(printf '%.2f' "$PCT")
-          if awk -v p="$PCT" -v t="$THRESHOLD" 'BEGIN{exit !(p >= t)}'; then
-            echo "PASS: aerospike-core line coverage is ${PCT_FMT}%, threshold is ${THRESHOLD}% -- coverage meets the threshold."
+          if awk -v p="$PCT" -v t="$COVERAGE_THRESHOLD" 'BEGIN{exit !(p >= t)}'; then
+            echo "PASS: aerospike-core line coverage is ${PCT_FMT}%, threshold is ${COVERAGE_THRESHOLD}% -- coverage meets the threshold."
           else
-            echo "FAIL: aerospike-core line coverage is ${PCT_FMT}%, threshold is ${THRESHOLD}% -- coverage is below the threshold."
+            echo "FAIL: aerospike-core line coverage is ${PCT_FMT}%, threshold is ${COVERAGE_THRESHOLD}% -- coverage is below the threshold."
           fi
-          cargo llvm-cov report -p aerospike-core --fail-under-lines "$THRESHOLD"
+          cargo llvm-cov report -p aerospike-core --fail-under-lines "$COVERAGE_THRESHOLD"
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -127,8 +127,20 @@ jobs:
             --no-default-features --features serialization,async,rt-tokio \
             --ignore-run-fail --no-report -- --test-threads=1
 
-      # Soft floor while coverage tooling beds in: current aerospike-core baseline
-      # is ~79% (recorded on CLIENT-5292). 75% leaves headroom to catch a real
-      # regression without blocking on day one; raise to 80% once confirmed.
+      # THRESHOLD is deliberately set above the real ~79% baseline (recorded on
+      # CLIENT-5292) right now, to confirm this gate actually fails when it
+      # should. Dial back to the real working floor (75%, raising to 80% once
+      # confirmed) once that's verified.
       - name: Coverage report (aerospike-core)
-        run: cargo llvm-cov report -p aerospike-core --fail-under-lines 75
+        run: |
+          THRESHOLD=90
+          cargo llvm-cov report -p aerospike-core
+          PCT=$(cargo llvm-cov report -p aerospike-core --json --summary-only | jq -r '.data[0].totals.lines.percent')
+          PCT_FMT=$(printf '%.2f' "$PCT")
+          if awk -v p="$PCT" -v t="$THRESHOLD" 'BEGIN{exit !(p >= t)}'; then
+            echo "PASS: aerospike-core line coverage is ${PCT_FMT}%, threshold is ${THRESHOLD}% -- coverage meets the threshold."
+            exit 0
+          else
+            echo "FAIL: aerospike-core line coverage is ${PCT_FMT}%, threshold is ${THRESHOLD}% -- coverage is below the threshold."
+            exit 1
+          fi
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -127,8 +127,7 @@ jobs:
           PCT_FMT=$(printf '%.2f' "$PCT")
           if awk -v p="$PCT" -v t="$THRESHOLD" 'BEGIN{exit !(p >= t)}'; then
             echo "PASS: aerospike-core line coverage is ${PCT_FMT}%, threshold is ${THRESHOLD}% -- coverage meets the threshold."
-            exit 0
           else
             echo "FAIL: aerospike-core line coverage is ${PCT_FMT}%, threshold is ${THRESHOLD}% -- coverage is below the threshold."
-            exit 1
           fi
+          cargo llvm-cov report -p aerospike-core --fail-under-lines "$THRESHOLD"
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test.yml
```diff
@@ -69,3 +69,66 @@ jobs:
           # aerospike-core doc examples use `use aerospike::…`; enable rt-tokio on the path dev-dep.
           sed -i 's/features = \["async"\]/features = ["async", "rt-tokio"]/' aerospike-core/Cargo.toml
           cargo test --doc -p aerospike-core --features rt-tokio,serialization
+
+  aerospike-core-coverage:
```
dmishra-aero, line 73: Why we have a separate job instead of folding this one into the `test` job ?

.github/workflows/build-test.yml
```diff
@@ -69,3 +69,66 @@ jobs:
...
+          PROPTEST_MAX_SHRINK_ITERS: 0
+        run: |
+          cargo llvm-cov -p aerospike \
+            --no-default-features --features serialization,async,rt-tokio \
+            --ignore-run-fail --no-report -- --test-threads=1
+
+      # 75% is the starting point to keep infra adoption moving -- nothing less.
+      - name: Coverage report (aerospike-core)
+        run: |
+          THRESHOLD=75
```
dmishra-aero: Can we use a Vars instead

.github/workflows/build-test.yml
```diff
@@ -69,3 +69,66 @@ jobs:
...
+            --no-default-features --features serialization,rt-tokio \
+            --ignore-run-fail --no-report
+
+      - name: Coverage - aerospike integration and property tests
+        env:
+          PROPTEST_MAX_SHRINK_ITERS: 0
+        run: |
+          cargo llvm-cov -p aerospike \
+            --no-default-features --features serialization,async,rt-tokio \
+            --ignore-run-fail --no-report -- --test-threads=1
```
dmishra-aero, line 124: Is the coverage report used as a Quality Gate for CI build  or the intent is just collect coverage irrespective of test failures

.github/workflows/build-test.yml
```diff
@@ -69,3 +69,66 @@ jobs:
...
+        uses: reugn/github-action-aerospike@2065a9209cfd5ef88a3e07f3e7929e321d1e0067 # v1.1.0
+
+      - name: Install cargo-llvm-cov
+        uses: taiki-e/install-action@1ed6d7be6168f6c9046541087ff549b6bc581fdf # v2.87.2
+        with:
+          tool: cargo-llvm-cov
+
+      - name: Coverage - aerospike-core unit tests
+        env:
+          PROPTEST_MAX_SHRINK_ITERS: 0
```
dmishra-aero, line 111: should this match with `test` Job  or was it done on purpose for speed sake

.github/workflows/build-test.yml
```diff
@@ -69,3 +69,66 @@ jobs:
...
+          cargo llvm-cov -p aerospike \
+            --no-default-features --features serialization,async,rt-tokio \
+            --ignore-run-fail --no-report -- --test-threads=1
+
+      # 75% is the starting point to keep infra adoption moving -- nothing less.
+      - name: Coverage report (aerospike-core)
+        run: |
+          THRESHOLD=75
+          cargo llvm-cov report -p aerospike-core
+          PCT=$(cargo llvm-cov report -p aerospike-core --json --summary-only | jq -r '.data[0].totals.lines.percent')
```
dmishra-aero, line 129: `cargo llvm-cov report -p aerospike-core --fail-under-lines <coverage_threshold> ?`
Instead of reading the score twice and writing your own `outcome` rule, use the tool’s built-in pass/fail.

dmishra-aero (changes requested):

.github/workflows/build-test.yml
diparthaspike: Two reasons:                                                                                                                                                                                             
                                                                                                                                                                                                           
  1. Clear failure signal. test (rt-tokio) / test (rt-async-std) and aerospike-core-coverage are two different concerns — one asks "do tests pass," the other asks "is the coverage floor met." If coverage
     were a step inside the test job, a red X could mean either, and you'd have to open the job and scroll through steps to tell which. As a separate job, the check name itself tells you what failed.    
  2. No added wall-clock cost. GitHub Actions runs jobs in parallel by default. As a separate job, coverage runs concurrently with the two test legs instead of stacking extra minutes onto rt-tokio's     
     critical path.                                                                                                                                                                                        
                                                                                                                                                                                                           
  The tradeoff, which we considered: this does mean the integration/proptest suite runs one extra time (once instrumented for coverage, on top of the existing test (rt-tokio) run), since coverage needs  
  its own differently-compiled binary. Given this is a public repo (free, unlimited Actions minutes on standard runners) and the extra job runs in parallel rather than serially, that cost is negligible  
  against the clarity gain.

.github/workflows/build-test.yml
diparthaspike: On purpose, for speed — proptest's default is up to 1024 shrink iterations per failure, which is what blew a local run out to 61 minutes before we added this. Fair point that test doesn't have the same
  guard though — it'd hit the exact same runaway-shrink problem if a proptest test failed there.

.github/workflows/build-test.yml
diparthaspike: addressed.

.github/workflows/build-test.yml
diparthaspike: Keeping it hardcoded for now — the 75% is meant to be a deliberate, reviewed number (with the comment explaining why right above it), and moving it to a repo variable means anyone with settings access 
  could quietly bump it without a PR, so the comment goes stale and nobody notices. Once we're ready to raise it to 80%, that'll go through a PR too. Happy to switch to vars later if we want ops to tune 
  it without code review.

.github/workflows/build-test.yml
diparthaspike: Yes, it's a real quality gate — coverage below 75% fails the aerospike-core-coverage check, same as any other required status. The intent isn't "collect coverage no matter what, don't judge it."       
                                                                                                                                                                                                           
  --ignore-run-fail is there for a narrower reason: without it, cargo-llvm-cov refuses to produce any report at all if even one test fails — you get a blank/zeroed-out coverage number instead of a real  
  one, hiding the actual signal. Test correctness is already enforced separately by the test job (untouched, still fails on any test failure) — this job's only job is "is coverage above the floor," and  
  it shouldn't go dark just because an unrelated test hiccupped elsewhere.

.github/workflows/build-test.yml
dmishra-aero: If you wanted to be versioned and track what had changed..then declare a env constant and use that varianle instead of hardcoding in steps.. 
But i feel var could be wise move since we are keep on changing it until we come to a correct threshold, even not everyone will have access to setting and that is being tracked when someone update's it..

dmishra-aero (approved):

khaf:
If the tests are run against a community server, we do not get any coverage on a number of files which will drive the avg coverage down considerably. What's the plan there?

diparthaspike:
> If the tests are run against a community server, we do not get any coverage on a number of files which will drive the avg coverage down considerably. What's the plan there?

@khaf 
Thanks for flagging this — good catch, this is an angle we hadn't fully thought through.

You're right: our current CI setup pulls the Community Edition image from Docker Hub, so Enterprise-only features like multi-record transactions and strong consistency can't actually be exercised against it. That means a meaningful chunk of the txn cluster (`txn_monitor.rs`, `txn_roll.rs`, the `txn_*_command.rs` files) can't get real integration coverage under the current setup, no matter how much we improve the tests.

Given that, we'd like to keep this PR scoped purely to introducing the coverage tooling itself, and lower the threshold to 60% for now so it doesn't block ongoing work while we figure out the Enterprise piece. As a fast follow-up, we'll open a separate ticket to bring in Enterprise Edition testing using the shared workflows setup (similar to what the Python client already does), and then write the missing tests for the Enterprise-only features once that's in place.

Attaching the current aerospike-core coverage breakdown below for reference.

| File | Lines | Missed Lines | Line Cover |
| :--- | :--- | :--- | :--- |
| batch/batch_executor.rs | 364 | 38 | **89.56%** |
| batch/batch_record.rs | 37 | 0 | **100.00%** |
| batch/mod.rs | 551 | 38 | **93.10%** |
| bin.rs | 26 | 7 | **73.08%** |
| client.rs | 1148 | 444 | **61.32%** |
| cluster/mod.rs | 1298 | 421 | **67.57%** |
| cluster/node.rs | 933 | 140 | **84.99%** |
| cluster/node_validator.rs | 198 | 92 | **53.54%** |
| cluster/partition.rs | 252 | 68 | **73.02%** |
| cluster/partition_tokenizer.rs | 89 | 26 | **70.79%** |
| cluster/peers.rs | 87 | 27 | **68.97%** |
| cluster/peers_parser.rs | 287 | 23 | **91.99%** |
| cluster/version_parser.rs | 137 | 9 | **93.43%** |
| commands/admin_command.rs | 806 | 784 | **2.73%** |
| commands/as_msg_fields.rs | 73 | 7 | **90.41%** |
| commands/batch_attr.rs | 185 | 48 | **74.05%** |
| commands/batch_operate_command.rs | 561 | 152 | **72.91%** |
| commands/buffer.rs | 2777 | 504 | **81.85%** |
| commands/delete_command.rs | 92 | 6 | **93.48%** |
| commands/execute_udf_command.rs | 75 | 8 | **89.33%** |
| commands/exists_command.rs | 91 | 9 | **90.11%** |
| commands/info_command.rs | 67 | 5 | **92.54%** |
| commands/mod.rs | 82 | 4 | **95.12%** |
| commands/operate_command.rs | 76 | 11 | **85.53%** |
| commands/particle_type.rs | 41 | 24 | **41.46%** |
| commands/query_command.rs | 74 | 6 | **91.89%** |
| commands/query_explain_command.rs | 183 | 163 | **10.93%** |
| commands/read_command.rs | 185 | 1 | **99.46%** |
| commands/server_command.rs | 217 | 53 | **75.58%** |
| commands/single_command.rs | 329 | 96 | **70.82%** |
| commands/stream_command.rs | 316 | 54 | **82.91%** |
| commands/touch_command.rs | 89 | 9 | **89.89%** |
| commands/txn_add_keys_command.rs | 93 | 93 | **0.00%** |
| commands/txn_close_command.rs | 62 | 62 | **0.00%** |
| commands/txn_mark_roll_forward_command.rs | 64 | 64 | **0.00%** |
| commands/txn_roll_command.rs | 70 | 70 | **0.00%** |
| commands/txn_verify_command.rs | 66 | 66 | **0.00%** |
| commands/write_command.rs | 102 | 0 | **100.00%** |
| common.rs | 6 | 0 | **100.00%** |
| errors.rs | 692 | 96 | **86.13%** |
| expressions/ael.rs | 64 | 0 | **100.00%** |
| expressions/bitwise.rs | 302 | 3 | **99.01%** |
| expressions/hll.rs | 144 | 30 | **79.17%** |
| expressions/lists.rs | 528 | 19 | **96.40%** |
| expressions/maps.rs | 659 | 61 | **90.74%** |
| expressions/mod.rs | 1486 | 149 | **89.97%** |
| expressions/string.rs | 414 | 414 | **0.00%** |
| key.rs | 201 | 30 | **85.07%** |
| mapping/mod.rs | 284 | 101 | **64.44%** |
| mapping/serde.rs | 528 | 185 | **64.96%** |
| mapping/serde_helpers.rs | 182 | 55 | **69.78%** |
| metrics/cluster.rs | 27 | 0 | **100.00%** |
| metrics/histogram.rs | 248 | 2 | **99.19%** |
| metrics/node_metrics.rs | 947 | 68 | **92.82%** |
| metrics/policy.rs | 147 | 0 | **100.00%** |
| msgpack/decoder.rs | 167 | 54 | **67.66%** |
| msgpack/encoder.rs | 582 | 87 | **85.05%** |
| net/buffer_pool.rs | 209 | 9 | **95.69%** |
| net/connection.rs | 1315 | 445 | **66.16%** |
| net/connection_pool.rs | 370 | 44 | **88.11%** |
| net/host.rs | 88 | 14 | **84.09%** |
| net/parser.rs | 104 | 3 | **97.12%** |
| operations/bitwise.rs | 403 | 25 | **93.80%** |
| operations/cdt.rs | 24 | 8 | **66.67%** |
| operations/cdt_context.rs | 453 | 120 | **73.51%** |
| operations/exp.rs | 80 | 26 | **67.50%** |
| operations/hll.rs | 203 | 23 | **88.67%** |
| operations/lists.rs | 894 | 57 | **93.62%** |
| operations/maps.rs | 967 | 32 | **96.69%** |
| operations/mod.rs | 76 | 7 | **90.79%** |
| operations/path.rs | 47 | 3 | **93.62%** |
| operations/scalar.rs | 72 | 0 | **100.00%** |
| operations/string.rs | 317 | 251 | **20.82%** |
| policy/admin_policy.rs | 5 | 0 | **100.00%** |
| policy/batch_policy.rs | 20 | 3 | **85.00%** |
| policy/client_policy.rs | 239 | 23 | **90.38%** |
| policy/expiration.rs | 7 | 0 | **100.00%** |
| policy/mod.rs | 113 | 24 | **78.76%** |
| policy/query_policy.rs | 48 | 3 | **93.75%** |
| policy/read_policy.rs | 25 | 0 | **100.00%** |
| policy/read_touch_ttl_percent.rs | 6 | 0 | **100.00%** |
| policy/txn_policy.rs | 47 | 6 | **87.23%** |
| policy/write_policy.rs | 39 | 0 | **100.00%** |
| privilege.rs | 83 | 83 | **0.00%** |
| query/filter.rs | 547 | 26 | **95.25%** |
| query/index_types.rs | 28 | 0 | **100.00%** |
| query/node_partitions.rs | 24 | 0 | **100.00%** |
| query/partition_filter.rs | 60 | 4 | **93.33%** |
| query/partition_status.rs | 20 | 0 | **100.00%** |
| query/partition_tracker.rs | 377 | 121 | **67.90%** |
| query/plan/index_range_wire.rs | 193 | 27 | **86.01%** |
| query/plan/query_plan.rs | 242 | 34 | **85.95%** |
| query/plan/query_where_wire.rs | 215 | 26 | **87.91%** |
| query/recordset.rs | 151 | 12 | **92.05%** |
| query/statement.rs | 55 | 8 | **85.45%** |
| query/udf.rs | 12 | 5 | **58.33%** |
| record.rs | 82 | 1 | **98.78%** |
| result_code.rs | 364 | 95 | **73.90%** |
| sampler.rs | 105 | 1 | **99.05%** |
| server_error.rs | 652 | 43 | **93.40%** |
| task/drop_index_task.rs | 42 | 2 | **95.24%** |
| task/execute_task.rs | 176 | 121 | **31.25%** |
| task/index_task.rs | 72 | 16 | **77.78%** |
| task/register_task.rs | 28 | 0 | **100.00%** |
| task/task.rs | 23 | 0 | **100.00%** |
| task/udf_remove_task.rs | 38 | 1 | **97.37%** |
| txn.rs | 207 | 160 | **22.71%** |
| txn_monitor.rs | 110 | 110 | **0.00%** |
| txn_roll.rs | 376 | 376 | **0.00%** |
| value.rs | 871 | 229 | **73.71%** |
| xor_shift.rs | 82 | 6 | **92.68%** |
| **TOTAL** | **31,197** | **7,619** | **75.58%** |

Tickets: 
- https://aerospike.atlassian.net/browse/CLIENT-5402 
- https://aerospike.atlassian.net/browse/CLIENT-5403

dmishra-aero:
Hi @diparthaspike ,
I have added a new commit..not validated properly, please try to validate and feel free to fix if something is not correct..

## Changes Addressed:

 **Replace aerospike-core-coverage with merge-coverage: all three test legs**
  (test, test-security, test-ap) emit lcov artifacts that are merged and
  gated at 60% with a warning summary — no PR block on dip
- Add test-ap (EE+AP mode, both runtimes) as a new PR-blocking job
- Promote test-security to full matrix (rt-tokio + rt-async-std)
- Drop --ignore-run-fail; coverage gate only runs when all tests pass

**Test helpers:**
- Split no_tls() into no_server_tls() + no_mutual_tls() for precise guards
- Update tls_client_no_auth → no_server_tls(), tls_client_auth → no_mutual_tls()
  (PKI test auto-activates once shared-workflows gains enable-tls-client-auth)

**Nightly:**
- nightly-branch.yml: reusable 5-leg workflow parameterised by inputs.branch
  (EE+SC, EE+TLS+security, EE+AP, CE, lua compile)
- nightly-v2.yml: cron 02:00 UTC → v2
- nightly-v3.yml: cron 02:30 UTC → v3


## Planned follow-ups: 

1. **Mutual TLS / PKI** — server infrastructure gap The Rust client code for **AuthMode::PKI** and `tls_client_auth` is complete and guarded by `no_mutual_tls().` CI cannot exercise it today because the `setup-aerospike-server` action always starts the server with `tls-authenticate-client false `— it has no input to enable mutual TLS and no `tls-mutual.conf `template. Until that action gains enable-tls-client-auth, the server will never demand a client certificate and the test will never run.

2. **External / LDAP auth** — no setup action exists anywhere **AuthMode::External** and ExternalInsecure require Aerospike to be fronted by an LDAP directory. No setup-ldap-server action exists in shared-workflows today, and no other client (Java, C#) tests this in CI either — this needs new infrastructure built from scratch before any client can cover these paths.

So, **Point 1** could be targeted as a new PR/Story.
For **Point 2**, we would need to check with the Shared Workflow team. It’s not something we urgently need to support, but it could be a useful addition to have going forward.

@khaf FYI...

khaf:
Thanks @dmishra-aero , I'll follow that up with @justinlee-aerospike to see how to push this issue forward.

@diparthaspike Please update this ticket with all the cases that are covered and merged, and the remaining cases requiring future work before the final review.

diparthaspike:
Flagging a dependency change worth a closer look: mlua is now pinned to 0.11 instead of 0.12.

mlua 0.12.x declares rust-version 1.88, which breaks this project's pinned 1.87 MSRV, so cargo build --features lua could no longer resolve dependencies at all (caught by the new lua-feature-build nightly job). 0.11.6 supports the same lua54, vendored, async, and send features and only needs rustc 1.85, so it drops in cleanly under our MSRV.

One follow-on change: mlua::LuaString isn't a top-level re-export in 0.11, only available via mlua::prelude, so the two call sites in aerospike-core/src/lua/bytes.rs were updated to use mlua::prelude::LuaString instead.

Verified locally: clean build under rustc 1.87, and all 474 aerospike-core unit tests pass with the lua feature enabled.

diparthaspike:
@dmishra-aero @khaf summary of what's in this PR now:                                                                                                                                                    
                                                                                                                                                                                                           
Coverage CI is split into 4 jobs (unit-test-coverage, test-ce, test-ee-security, test-ee-ap) plus a merge-coverage gate, all green at 83.50% coverage. Renamed jobs so they actually reflect what they test (test-ce/test-ee-security/test-ee-ap instead of the old test-sc misnomer). Fixed a nightly TLS leg bug where the whole suite was hitting a TLS-only port instead of just the TLS-aware tests. Pinned mlua to 0.11, since 0.12 needs rustc 1.88 and breaks our 1.87 MSRV pin. Dropped v2 nightly work from this PR entirely, since it's opened up its own scope (TTL guards, test bugs, DB version questions) will track that separately.

dmishra-aero (approved):
Look's fine

khaf (approved):
Looks good.

PR #254, opened 2026-09-04
[CLIENT-5402]: use enterprise db in github actions

diparthaspike:
Fixes: https://aerospike.atlassian.net/browse/CLIENT-5402

Switch CI's Aerospike server from Community Edition to Enterprise Edition with strong consistency enabled, using the org's shared aerospike/shared-workflows setup-aerospike-server action instead of reugn/github-action-aerospike, in both the test job and the aerospike-core-coverage job.

Why: tests/src/txn.rs already has 18 MRT/SC integration tests that were silently skipping in CI (skip_if_no_mrt macro), since Community Edition can't support strong consistency. This PR gets them actually running instead of never executing at all.

Turning this on surfaced three tests that broke once they ran for real, all fixed here:
- batch_sc_delete_non_durable_forbidden_when_record_exists now branches on the live strong-consistency-allow-expunge setting (via a new ServerCapabilities.sc_allow_expunge field) instead of assuming one server config.
- example_record_operations now skips cleanly when explicit record TTL isn't allowed on the namespace, matching the existing ServerCapabilities pattern used elsewhere.
- batch_sc_delete_generation_mismatch_errors was checking the wrong thing: client.batch() only fails its outer Result for whole-batch errors, per-key outcomes live in each BatchRecord.result_code. The server was already correctly rejecting the write; the test assertion was wrong, not the client.

Result: aerospike-core line coverage is now 80.79%, up from the 75.58% Community-only baseline.

Requires new AEROSPIKE_SERVER_FEATURES_CONF secret (added for this PR).

Verified here: https://github.com/aerospike/aerospike-client-rust/actions/runs/33845519696

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -7,6 +7,10 @@ on:
       - v3
   workflow_dispatch:
 
+permissions:
+  contents: read
+  id-token: write  # required by setup-aerospike-server's JFrog OIDC login
+
 env:
   CARGO_TERM_COLOR: always
   RUST_BACKTRACE: 1
@@ -44,7 +48,12 @@ jobs:
           shared-key: ${{ matrix.runtime }}
 
       - name: Set up Aerospike Database
-        uses: reugn/github-action-aerospike@2065a9209cfd5ef88a3e07f3e7929e321d1e0067 # v1.1.0
+        uses: aerospike/shared-workflows/.github/actions/setup-aerospike-server@0d843cff870d70ac527048ab0d006c7c301fbb2f # v4.1.0
+        with:
+          enable-strong-consistency: "true"
+          features-content: ${{ secrets.AEROSPIKE_SERVER_FEATURES_CONF }}
+          oidc-provider: ${{ vars.JFROG_OIDC_PROVIDER }}
+          oidc-audience: ${{ vars.JFROG_OIDC_AUDIENCE }}
 
       - name: Build workspace
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/build-test.yml
```diff
@@ -44,7 +48,12 @@ jobs:
           shared-key: ${{ matrix.runtime }}
 
       - name: Set up Aerospike Database
-        uses: reugn/github-action-aerospike@2065a9209cfd5ef88a3e07f3e7929e321d1e0067 # v1.1.0
+        uses: aerospike/shared-workflows/.github/actions/setup-aerospike-server@0d843cff870d70ac527048ab0d006c7c301fbb2f # v4.1.0
```
dmishra-aero, line 51: Currently this will use `latest` server tag and future bump can change CI behavior without any Rust code changing so pinning to specific server tag(may be last previous released tag) would make our workflow more deterministic to failures..

dmishra-aero (approved):
Look's fine

.github/workflows/build-test.yml
diparthaspike: Agreed, is there a single server version we're all supposed to pin and bump together, or does each client repo just do its own thing? I checked a few sibling repos and the pinned versions are all over the place (Go is on 8.1.2.0, Java has 8.1.0.3_2/8.1.1.0_1, Node uses 8.1.1.0), so I wasn't sure if there's a shared source of truth I missed. Either way, what version should we target for now?

dmishra-aero: `8.1.2.0` we could use this for now.. and then later we can extend it to other version too..

PR #257, opened 2026-09-11
CLIENT-5414 set TCP_NODELAY

gagan405:
Continuation from https://github.com/aerospike/aerospike-client-rust/pull/252


khaf (approved):
Looks good.

PR #259, opened 2026-09-13
CLIENT-5457: Return an abandoned roll-forward as CommitError so the cause is not dropped

paul-brackin:
Ok(RollForwardAbandoned) swallowed the timeout/server failure; JSDK keeps that error on the thrown Commit exception. CLOSE_ABANDONED is still success.


khaf (approved):
Look good.

PR #260, opened 2026-09-15
Remove rhexdump to avoid Apache 2.0 and GPL 3.0 friction

kjanosz:
[rhexdump](https://crates.io/crates/rhexdump/0.2.0) is licensed under GPL 3.0+. I am not a lawyer, but as far as I understand the Apache 2.0 license, which aerospike-client uses, is incompatible with GPL 3.0+. 
Whole dependency does not seem to be used anyway, except of one method with singular call-site commented out.

I detected it because it fails `cargo deny` checks when copyleft licenses are excluded ;-)


khaf (approved):
Looks good.

khaf:
Thank for letting us know.

